PR #35, opened 2024-03-18
Spanish translation

k2helix:


app/src/main/res/values-es/strings.xml
```diff
@@ -0,0 +1,65 @@
...
+    <string name="ok">OK</string>
+    <string name="cancel">Cancelar</string>
+    <string name="confirm">Confirmar</string>
+    <string name="are_you_sure_delete_medicine">¿Quieres eliminar el medicamento?</string>
+    <string name="no_reminders">Ningún recordatorio establecido</string>
+    <plurals name="sum_reminders">
+        <item quantity="one">%1$d recordatorio\n%2$s</item>
+        <item quantity="other">%1$d recordatorios\n%2$s</item>
+    </plurals>
+    <string name="channel_name" translatable="false">MedTimer</string>
```
Futsch1: Please remove the translations with the flag "translatable" set to false from the XML.

app/src/main/res/values-es/strings.xml
```diff
@@ -0,0 +1,65 @@
...
+    <string name="channel_description" translatable="false">Notificaciones de MedTimer</string>
```
Futsch1: Please remove the translations with the flag "translatable" set to false from the XML.

k2helix: You're right! I didn't notice them as I didn't use Android Studio to edit the file. Sorry!

Futsch1:
Thanks a lot for the PR, mucb appreciated!

k2helix:
No problem! Thank you for the app!

k2helix:
Hi @Futsch1, I noticed a typo in the medicine tab. I translated it as "Medicamento" which is a single medicine. The correct translation would be either "Medicamentos" (in plural) or "Medicación" (which means medication). Considering that the first one would require more space, I recommend the second one.

(The literal translation of medicine is "Medicina", and even though one of its meanings is the same, it's more used to refer to the profession/treatment, so it'd be confusing)

Futsch1:
Thanks for the hint, I'll fix it

PR #127, opened 2024-05-19
Add Inventory by Medecine

docjyJ:
- [x] Addition of the remaining quantity in the database.
- [ ] Decreed at each event.
- [ ] Personalize the number of medicine per event.
- [ ] Update translations.
- [ ] (Optional/discuss) Modify the UI of medicine.
- [ ] (Optional/discuss) Use set of colors instead of picker.
- [ ] (Optional/discuss) Add icon, pill, capsule, syrup, etc.

### DIFF
--- a/app/schemas/com.futsch1.medtimer.database.MedicineRoomDatabase/7.json
+++ b/app/schemas/com.futsch1.medtimer.database.MedicineRoomDatabase/7.json
@@ -1,0 +1,225 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 7,
+    "identityHash": "a11832724e19391d8fab1640c20a919e",
+    "entities": [
+      {
+        "tableName": "Medicine",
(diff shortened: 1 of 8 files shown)


Futsch1:
Hey Jean-Yves, thanks for your contribution and your willingness to improve the app. I have so far refrained from working on the inventory because the amount is (and it seems to make sense this way) a string and not a number. So you might have to check if the amount is actually a number in order to be able to use inventories.

So it's all a bit complicated and would also make the UI more complex. But if you find a nice solution to these issues I would be perfectly willing to add the feature (and the others planned by you) to the app.

docjyJ:
Hi,

Right now, I don't have enough time to work.

I still use your application is it is still as great!

It is true that this functionalty is commplex because of string.

Thank you for your work. :heart:

Omarwk:
Maybe adding the image of the med would help also the people who vision challenged

docjyJ:
Yes, I don't have enough time to work on it. Maybe work on a feature request.

PR #144, opened 2024-05-31
Add italian translation

### LINKED ISSUE #142: Italian translation
Hello again,
I saw there's values-de but there's no italian translation. 
How can I contribute? Is there an XML file I can easily edit locally and then upload to contribute the italian translation?
Thnak you very much!

estux:
Manually checked and edited the italian translation of the app.

This PR once merged will close #142 

3 notes:
- I tried to respect the format of the original english translation, talking about "\\n" and other escape characters. **Please double check the format and the spaces in the app before merging this PR** (so we can avoid strange simbols in the menus or in the text descriptions of the app, or also cutted descriptions or overflowed texts in buttons)
- I encountered some strings with "translatable=false" and I didn't translated the text inside, e.g. \<string name="sort" translatable="false"\>Sort\</string\>
- I tried to not use gender based nouns or verbs and reformulated a bit in those cases to use neutral and more direct form in italian

I would gladly help to translate also the Description of the app for the app stores like Play Store or F-Droid


Futsch1:
Nice work, thanks a lot!

PR #252, opened 2024-09-28
246 fr add icon per medication

Futsch1:


docjyJ:
Great! Your app is awesome :heart:

PR #393, opened 2024-12-20
Fix building APK

Futsch1:


wh201906:
Hi. I wonder if it's possible to make a new release for this commit or tag this as `v1.13.0`.

PR #631, opened 2025-05-09
Swedish translation

IdaEdz:
I really wanted my own language in this otherwise perfect app!

Changes:
+ Added Swedish translation for the app. (../values-sv/strings.xml)

Comment: It may require proofing and in-app testing. But this should be a good starting point.


IdaEdz:
I don't really understand these checks. Is there something wrong with my request? I did not change any code, only added a translation file. Do I need to do something more?

IdaEdz:
I'm going to redo this via Hosted Weblate...

Futsch1:
Hey @IdaEdz, sorry, I just started the checks yesterday but did not look into this yet. All that was wrong is a component not being up to date - not your fault!

Highly appreciating your contribution, thanks a lot!!! Using Weblate might be the easier way to contribute, so I'll tend to that MR.

PR #744, opened 2025-07-26
Wear Pair

tanquetav:
Initial support for wearos

c1abcba0-6da1-4b14-a930-59a4b050432c
1e95a786-0ec5-451c-8731-1da40472efbe


Futsch1:
Whoa, that's a big one! Thanks a lot for your contribution!

I will need some time to do a review, so have some patience (and you'll have to sort out the failing checks)...

But maybe before I start @tanquetav: What was the motivation to implement a separate wear app? Which added value does it provide compared to the notifications that are sent to the watch anyway?

tanquetav:
> Whoa, that's a big one! Thanks a lot for your contribution!
> 
> I will need some time to do a review, so have some patience (and you'll have to sort out the failing checks)...
> 
> But maybe before I start @tanquetav: What was the motivation to implement a separate wear app? Which added value does it provide compared to the notifications that are sent to the watch anyway?

I have problems with notification on android. I receive a lot of them then I disabled all of them. I implemented this solution to watch vibrate in a different pattern (1 full second) and the companion activity (second image) appears on foreground in the watch.  It works better for me.

Probably I will need to verify your pipelines if it makes sense. The strange is I did not change any screen in the Phone app and the e2e tests failed.

Futsch1:
@tanquetav Any intention of working on the failing tests or rebasing to the current main? If not, I'd be inclined to close the PR...

tanquetav:
Unfortunately I have no idea about the fails. It fails in several points I did not touch.
You can close the PR. I build my own version (sometimes I am rebasing with your code)

PR #810, opened 2025-08-29
Add the ability to add notes for medications

### LINKED ISSUE #626: [Feature Request] Add a Notes Section for Each Medication
I'm unfortunately dealing with gout, and one of the commonly used medications is colchicine. The dosage for colchicine varies significantly between daily maintenance and acute flare-ups (e.g., one tablet per day versus three tablets per day).
Currently, the system allows us to add "Additional Instructions" to each medication reminder. However, in my case, I need to create separate reminders for different situations, and copying the same detailed information (e.g., usage) into each reminder can be redundant.
It would be more efficient to have a dedicated Notes section at the medication level, where general information—such as usage guidelines, side effects, and official leaflet content—can be stored and referenced as needed.

Add a Notes field to each medication, so users can record general reference information independent of individual reminders.

Thank you so much for your hard work on this app. It has helped me get through many difficult times with my health, and I truly appreciate the care and thought that went into its design.

BlankEclair:
[s.webm](https://github.com/user-attachments/assets/c7da834c-2785-4709-a763-04d9347f4a5e)

Closes #626


BlankEclair:
Trying to restore a backup made with older versions fails; is this intentional?

Futsch1:
No, restoring an old backup should work - even though I have not thoroughly tested it in all edge and corner cases and have no automatic tests for it. 

Regarding the failing test: I hoped to have that fixed a week ago, but seems not to be the case. Shouldn't be caused by your changes.

I will try to review the change, but I am on holiday and only have my mobile phone to do it...

Thank you for the contribution!

app/build.gradle.kts
```diff
@@ -24,7 +24,7 @@ android {
         multiDexEnabled = true
         targetSdk = 36
         versionCode = 131
```
Futsch1: You also have to bump the version code at the same time, otherwise Android won't detect an actual change of the version.

Futsch1 (changes requested):
There was one more aspect in the issue: to show the notes in the medicine list. Have you considered that? 

And: Please add a short Android test to it.

Apart from this: great work, much appreciated!

BlankEclair:
> There was one more aspect in the issue: to show the notes in the medicine list. Have you considered that?

Do you mean the list of medicines in the UI, or the medicines export?

Futsch1:
> > There was one more aspect in the issue: to show the notes in the medicine list. Have you considered that?
> 
> Do you mean the list of medicines in the UI, or the medicines export?

I mean the list of medicines in the UI, that's what the requestor of the feature wanted if I got this right.

BlankEclair:
> What I’m hoping for is simply to have a Notes field shown within the medication view screen (as seen in [3.png](https://github.com/Futsch1/medTimer/blob/main/fastlane/metadata/android/en-US/images/phoneScreenshots/3.png)), so I can quickly refer to key information when adjusting reminders.

 _Originally posted by @undecV in [#626](https://github.com/Futsch1/medTimer/issues/626#issuecomment-2851313837)_

Also, I'm personally going to use notes to store observed side effects out of curiosity, and I don't really need to see it every day ^^;

Futsch1:
You are right, then let's keep it the way it is right now.

PR #857, opened 2025-09-27
Fix manual doses causing duplicate dosages

### LINKED ISSUE #856: Logging additional dose of a disabled reminder leads to duplicate dosages
**Describe the bug**
Logging an additional dose of a medication with a disabled reminder leads to an event that is titled "&lt;medication&gt; (&lt;dosage&gt;)", with a dosage of... the dosage

**To Reproduce**
1. Create a medicine
2. Create a reminder
3. Disable said reminder
4. Go to log an additional dose
5. Log a new dose of the disabled reminder

**Expected behavior**
The event be simply called the name of the medication, with the dosage being the dosage

**Screenshots**

https://github.com/user-attachments/assets/67b80881-7a2d-4b39-9e0b-b8d6f1632d92

**Smartphone (please complete the following information):**

- Device: Google Pixel 8a
- Android version: Android 15

BlankEclair:
Fixes #856

### DIFF
--- a/app/src/androidTest/java/com/futsch1/medtimer/ManualDoseTest.kt
+++ b/app/src/androidTest/java/com/futsch1/medtimer/ManualDoseTest.kt
@@ -67,4 +67,28 @@ class ManualDoseTest : BaseTestHelper() {
         clickOn(com.google.android.material.R.id.material_timepicker_ok_button)
     }
 
+    @Test
+    //@AllowFlaky(attempts = 1)
+    fun testManualDoseOfDisabledReminder() {
+        // Create medication + reminder
+        AndroidTestHelper.createMedicine("Test")
+        AndroidTestHelper.createReminder("1 pill", null)
+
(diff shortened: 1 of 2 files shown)


Futsch1:
Thanks a lot for your PR and the fix.

However, I need to check this why I added the baseName/name stuff in the ManualLogEntry in the first place. There was a reason for that. So please be patient until I can merge this, because I'd like to dig deeper a bit first!

BlankEclair:
Okay ^_^

(I also tried to figure it out, but gave up lol, so I'm delegating that 
part to you)

Futsch1:
Ok, did the research: Your fix is actually correct. The name field of the ManualLogEntry is used in the dropdown that shows the list of possible meds and amounts to be logged (here it needs to be "Med (Amount)". In the actual logged entry however we need only the baseName, since the amount is stored in a separate filed.

PR #876, opened 2025-10-09
Append cycle info to events only when active days is greater then 1

### LINKED ISSUE #873: Add option to disable appending cycle information to events
I have medication that is taken once every two days. To handle this in MedTimer, I have set both consecutive days and pause days to 1. However, events for that medication have a superfluous "(1/1)" appended to the name.

Add an option to "Add cycle information to reminder events", so that I can disable it for said medication.

BlankEclair:
Fixes #873


Futsch1:
Thank you for the fix and for extending the tests!

I would like to share one aspect regarding the test concept: the general idea is to try to test as much as possible using unit tests and use the Android tests for the holistic, overarching aspects. So your fix could have been tested using unit tests only, but it is also fine the way you did it now.

Issue #131, opened 2024-05-22
"Linked" medicine alarm times


Jakarrrg:
It would be very helpful if some medicines could be "linked". For example my father takes one drop from  medicine x at 8 pm, and then another drop from  medicine y, ten minutes after the medicine x (8:10). Now, in many occasions he snoozes the alarm for medicine x for let's say 30 minutes (>8:40). It would be very helpful if medicine y would also snooze automatically for 30 minutes (>8:50) retaining the 10 minute interval from medicine x.

 Also it would be nice if user could choose at the time the alarm rings, or even a few minutes before, for how much time he/she/it wants to snooze the alarm, instead of the preselected from settings snooze time.

Futsch1:
Linking medicines sounds like a fairly complex feature, I will have to think about if and if yes how to implement it.

Custom snooze times should be fairly easy, but I will create a new issue for it to track the progress.

Futsch1:
I decided to implement this feature in the release after the next. I plan to implement it as following:

On the Advanced Settings page of a reminder, there will be a new button to add a linked reminder. This reminder will then make it possible to set a delay time that the linked reminder will trigger after the source reminder. So when the source reminder is marked taken, the delay time for the linked reminder will start.

It will not be possible to set the delay time to a value that would span to the next day, but same day is possible.

This will facilitate the use case that two reminders are taken with a defined delay between them.

Arkonos:
Am I correct in my understanding that only the same medicine, although with a different dose, can be linked?

Futsch1:
That is correct. However, that is currently only a limitation of the UI (if I'm not mistaken), so being able to link doses across medicines could be implemented. If you have the use case, feel free to create a feature request.

PR #976, opened 2025-11-26
Allow linked reminders for different medicine

Arkonos:
This PR aims to allow the user to select a different medicine (B) for the linked reminder than (A) that of the parent reminder.
This would fulfil the original ask of #131 more closely.

When creating a linked reminder, the user will first be prompted to select one of the existing medicines from `medicineRepository.medicines`. This list includes all, including A, but does not offer the `custom` field. Then the dose and delay have to be entered as before.
As the user will still be at the medicine A and not see the new linked reminder, a toast message will inform about the creating of the same. The linked reminder can be seen edited in medicine B.

Since I am not familiar with Kotlin or Java, everything has been coded by AI so far.
I've tested it in an emulator and it works. Before investing more time, I wanted to ask if you're generally in favour of this approach.

Some questions for you
1. I wanted to keep the changes to a minimum so far, so I'm reusing ManualDose functionality. Since the list of existing medication is now used in two different places, do you want me to create a separate module instead?
2. Do you want an indicator in the parent reminder that a linked one exists?
3. So far, there's no warning before deleting a parent reminder with a link. Given that a relationship is now reaching across more than one view, do you think this is necessary? The same would apply to deleting medicine A as a whole.
4. Do you want me to change the toast message to a clickable one, leading the user to B? I don't know how much more complex that'd be.

My personal use case is actually the same medication, but in a different dose. Since there are two different pills, I need to track the stock separately. With this PR, I can create `Medication A 5 mg`  and `Medication A 10 mg` and use a linked reminder.
Having it under one medication would be cleaner, but would probably mean a lot of work.

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/medicine/LinkedReminderHandling.kt
+++ b/app/src/main/java/com/futsch1/medtimer/medicine/LinkedReminderHandling.kt
@@ -4,6 +4,10 @@ import android.content.Context
 import android.os.Handler
 import android.os.Looper
 import androidx.fragment.app.FragmentActivity
+import android.app.AlertDialog
+import android.content.DialogInterface
+import com.futsch1.medtimer.overview.ManualDoseListEntryAdapter
+import com.futsch1.medtimer.overview.ManualDose
 import com.futsch1.medtimer.R
 import com.futsch1.medtimer.database.MedicineRepository
 import com.futsch1.medtimer.database.Reminder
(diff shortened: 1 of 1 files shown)


Futsch1:
@Arkonos Thanks a lot for your contribution, and also thanks for disclosing that you used AI!

1. I like it that you used the existing code. In terms of clarity, it would be very welcome to create a separate module that hosts the logic to create medicine selection lists and that is then used by both use cases in the manual dose and the linked reminder.
2. This would be a great addition, but not necessary to merge the PR from my point of view.
3. Same here, this might be rather seen as an optimization. Maybe create an issue for 2. and 3. to implement this later/separately.
4. No, that is fine as it is.

One thing that keeps the PR from being merged is the fail of the tests, which are due to the additional UI interaction required when creating linked reminders. Please have a look at the androidTests that run there and fix them accordingly. Note: In order to run them locally, you need to set the date/time of the device to 1.8.25, 16:00 (or use the adb command `adb shell su 0 toybox date 0801160025`). If you have issues getting the tests to work, let me know and I can support you.

PR #1249, opened 2026-03-04
Feature/1234 analysis view jetpack compose

ThomasKiljanczykDev:


Futsch1:
Phew, that's a big one. I will review it, but I would actually like to merge it. If we don't, I suppose all the work you put into it will be gone, and it will also be hard to keep it in sync with main if I continue implementing features.

What we should have is working tests, and, as I said, I will review. Is there anything else I can do to help? Some refactoring on main? And if you like, we can discuss potential architectural changes together. I know that the architecture is not great, the app started as a pure Java app with my Android knowledge from ten years ago and has been growing a lot since...

ThomasKiljanczykDev:
I'm closing this one for now

PR #1262, opened 2026-03-06
Convert database classes to kotlin

Futsch1:
@ThomasKiljanczykDev This is the first step in the Kotlin conversion

I integrated some of your review comments already. Removing @JvmField caused plenty of (non-logic affecting) changes in the unit tests, but this is basically a consequence of the small-step migration.

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/database/FullMedicine.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/FullMedicine.kt
@@ -4,6 +4,7 @@ import androidx.room.Embedded
 import androidx.room.Junction
 import androidx.room.Relation
 import com.google.gson.annotations.Expose
+import java.util.Objects
 
 class FullMedicine {
     @Embedded
@@ -19,25 +20,21 @@ class FullMedicine {
     var reminders: MutableList<Reminder> = mutableListOf()
 
     override fun equals(other: Any?): Boolean {
-        if (other == null || javaClass != other.javaClass) return false
-        val that = other as FullMedicine
-        return medicine == that.medicine && reminders == that.reminders && tags == that.tags
+        if (other !is FullMedicine) return false
+        return medicine == other.medicine && reminders == other.reminders && tags == other.tags
     }
 
     override fun hashCode(): Int {
-        var result = medicine.hashCode()
-        result += reminders.hashCode()
-        result += tags.hashCode()
-        return result
+        return Objects.hash(medicine, reminders, tags)
     }
 
     val isOutOfStock: Boolean
-        get() = this.isStockManagementActive && reminders.any { reminder: Reminder -> reminder.reminderType == Reminder.ReminderType.OUT_OF_STOCK && reminder.outOfStockThreshold >= medicine.amount }
+        get() = this.isStockManagementActive && reminders.any { reminder -> reminder.reminderType == Reminder.ReminderType.OUT_OF_STOCK && reminder.outOfStockThreshold >= medicine.amount }
 
     val isStockManagementActive: Boolean
         get() = (medicine.amount != 0.0 || hasStockReminder())
 
     private fun hasStockReminder(): Boolean {
-        return reminders.any { reminder: Reminder? -> reminder!!.reminderType == Reminder.ReminderType.OUT_OF_STOCK }
+        return reminders.any { reminder -> reminder.reminderType == Reminder.ReminderType.OUT_OF_STOCK }
     }
 }
--- a/app/src/main/java/com/futsch1/medtimer/database/JSONMedicineBackup.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/JSONMedicineBackup.kt
@@ -6,10 +6,10 @@ import java.time.Instant
 
 class JSONMedicineBackup : JSONBackup<FullMedicine>(FullMedicine::class.java) {
     override fun createBackup(databaseVersion: Int, list: List<FullMedicine>): JsonElement {
-        // Fix the medicines where the instructions are null
+        // Correct the medicines where the reminder's instructions are null in this loop
         for (fullMedicine in list) {
-            fullMedicine.reminders.filter { reminder: Reminder? -> reminder!!.instructions == null }
-                .forEach { reminder: Reminder? -> reminder!!.instructions = "" }
+            fullMedicine.reminders.filter { reminder -> reminder.instructions == null }
+                .forEach { reminder -> reminder.instructions = "" }
         }
         return super.createBackup(databaseVersion, list)
     }
--- a/app/src/main/java/com/futsch1/medtimer/database/Medicine.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/Medicine.kt
@@ -83,20 +83,20 @@ class Medicine() {
     }
 
     override fun equals(other: Any?): Boolean {
-        if (other == null || javaClass != other.javaClass) return false
-        return membersEqual(other as Medicine)
+        if (other !is Medicine) return false
+        return membersEqual(other)
     }
 
     override fun hashCode(): Int {
         return Objects.hash(medicineId, name, useColor, color, notificationImportance, iconId, amount, refillSizes, unit, notes, showNotificationAsAlarm)
     }
 
-    private fun membersEqual(that: Medicine): Boolean {
-        return medicineId == that.medicineId &&
-                name == that.name && useColor == that.useColor && color == that.color && notificationImportance == that.notificationImportance && iconId == that.iconId && amount == that.amount &&
-                refillSizes == that.refillSizes &&
-                unit == that.unit &&
-                notes == that.notes && showNotificationAsAlarm == that.showNotificationAsAlarm && expirationDate == that.expirationDate && productionDate == that.productionDate
+    private fun membersEqual(other: Medicine): Boolean {
+        return medicineId == other.medicineId &&
+                name == other.name && useColor == other.useColor && color == other.color && notificationImportance == other.notificationImportance && iconId == other.iconId && amount == other.amount &&
+                refillSizes == other.refillSizes &&
+                unit == other.unit &&
+                notes == other.notes && showNotificationAsAlarm == other.showNotificationAsAlarm && expirationDate == other.expirationDate && productionDate == other.productionDate
     }
 
     val refillSize: Double
--- a/app/src/main/java/com/futsch1/medtimer/database/MedicineDao.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/MedicineDao.kt
@@ -136,7 +136,7 @@ interface MedicineDao {
     @Query("SELECT COUNT(*) FROM Tag")
     fun countTags(): Int
 
-    @get:Query("SELECT MAX(sortOrder) FROM Medicine")
+    @get:Query("SELECT COALESCE(MAX(sortOrder), 0) FROM Medicine")
     val highestMedicineSortOrder: Double
 
     @Query("SELECT * FROM ReminderEvent WHERE reminderId= :reminderId ORDER BY remindedTimestamp DESC")
--- a/app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
@@ -9,6 +9,7 @@ import java.util.LinkedList
 import java.util.concurrent.Callable
 import java.util.concurrent.ExecutionException
 
+// TODO: Context should not be a property, but ReminderNotificationProcessor.buildReminderEvent must be changed before this is possible
 open class MedicineRepository(val context: Context) {
     private val medicineDao: MedicineDao
     private val database: MedicineRoomDatabase = MedicineRoomDatabase.getDatabase(context)
--- a/app/src/main/java/com/futsch1/medtimer/database/Reminder.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/Reminder.kt
@@ -149,8 +149,8 @@ class Reminder(var medicineRelId: Int) {
         }
 
     override fun equals(other: Any?): Boolean {
-        if (other == null || javaClass != other.javaClass) return false
-        return membersEqual(other as Reminder)
+        if (other !is Reminder) return false
+        return membersEqual(other)
     }
 
     override fun hashCode(): Int {
@@ -183,12 +183,12 @@ class Reminder(var medicineRelId: Int) {
         )
     }
 
-    private fun membersEqual(that: Reminder): Boolean {
-        return reminderId == that.reminderId && medicineRelId == that.medicineRelId && timeInMinutes == that.timeInMinutes && createdTimestamp == that.createdTimestamp && consecutiveDays == that.consecutiveDays && pauseDays == that.pauseDays &&
-                instructions == that.instructions && cycleStartDay == that.cycleStartDay &&
-                amount == that.amount &&
-                days == that.days && active == that.active && periodStart == that.periodStart && periodEnd == that.periodEnd && activeDaysOfMonth == that.activeDaysOfMonth && linkedReminderId == that.linkedReminderId && intervalStart == that.intervalStart && intervalStartsFromProcessed == that.intervalStartsFromProcessed && variableAmount == that.variableAmount && intervalStartTimeOfDay == that.intervalStartTimeOfDay && intervalEndTimeOfDay == that.intervalEndTimeOfDay && windowedInterval == that.windowedInterval &&
-                outOfStockThreshold == that.outOfStockThreshold && outOfStockReminderType == that.outOfStockReminderType && expirationReminderType == that.expirationReminderType
+    private fun membersEqual(other: Reminder): Boolean {
+        return reminderId == other.reminderId && medicineRelId == other.medicineRelId && timeInMinutes == other.timeInMinutes && createdTimestamp == other.createdTimestamp && consecutiveDays == other.consecutiveDays && pauseDays == other.pauseDays &&
+                instructions == other.instructions && cycleStartDay == other.cycleStartDay &&
+                amount == other.amount &&
+                days == other.days && active == other.active && periodStart == other.periodStart && periodEnd == other.periodEnd && activeDaysOfMonth == other.activeDaysOfMonth && linkedReminderId == other.linkedReminderId && intervalStart == other.intervalStart && intervalStartsFromProcessed == other.intervalStartsFromProcessed && variableAmount == other.variableAmount && intervalStartTimeOfDay == other.intervalStartTimeOfDay && intervalEndTimeOfDay == other.intervalEndTimeOfDay && windowedInterval == other.windowedInterval &&
+                outOfStockThreshold == other.outOfStockThreshold && outOfStockReminderType == other.outOfStockReminderType && expirationReminderType == other.expirationReminderType
     }
 
     enum class ReminderType(@param:DrawableRes val icon: Int) {
--- a/app/src/main/java/com/futsch1/medtimer/database/ReminderEvent.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/ReminderEvent.kt
@@ -81,8 +81,8 @@ class ReminderEvent {
         get() = reminderType == ReminderType.OUT_OF_STOCK || reminderType == ReminderType.EXPIRATION_DATE || reminderType == ReminderType.REFILL
 
     override fun equals(other: Any?): Boolean {
-        if (other == null || javaClass != other.javaClass) return false
-        return membersEqual(other as ReminderEvent)
+        if (other !is ReminderEvent) return false
+        return membersEqual(other)
     }
 
     override fun hashCode(): Int {
@@ -108,12 +108,12 @@ class ReminderEvent {
         )
     }
 
-    private fun membersEqual(o: ReminderEvent): Boolean {
-        return reminderEventId == o.reminderEventId &&
-                medicineName == o.medicineName &&
-                amount == o.amount && color == o.color && useColor == o.useColor && status == o.status && remindedTimestamp == o.remindedTimestamp && processedTimestamp == o.processedTimestamp && reminderId == o.reminderId && notificationId == o.notificationId && iconId == o.iconId && remainingRepeats == o.remainingRepeats && stockHandled == o.stockHandled && askForAmount == o.askForAmount &&
-                tags == o.tags && lastIntervalReminderTimeInMinutes == o.lastIntervalReminderTimeInMinutes &&
-                notes == o.notes && reminderType == o.reminderType
+    private fun membersEqual(other: ReminderEvent): Boolean {
+        return reminderEventId == other.reminderEventId &&
+                medicineName == other.medicineName &&
+                amount == other.amount && color == other.color && useColor == other.useColor && status == other.status && remindedTimestamp == other.remindedTimestamp && processedTimestamp == other.processedTimestamp && reminderId == other.reminderId && notificationId == other.notificationId && iconId == other.iconId && remainingRepeats == other.remainingRepeats && stockHandled == other.stockHandled && askForAmount == other.askForAmount &&
+                tags == other.tags && lastIntervalReminderTimeInMinutes == other.lastIntervalReminderTimeInMinutes &&
+                notes == other.notes && reminderType == other.reminderType
     }
 
     enum class ReminderStatus {
--- a/app/src/test/java/com/futsch1/medtimer/MedicineHelperTest.kt
+++ b/app/src/test/java/com/futsch1/medtimer/MedicineHelperTest.kt
@@ -86,7 +86,7 @@ class MedicineHelperTest {
         val medicine = Medicine("test")
         val fullMedicine = FullMedicine()
         fullMedicine.medicine = medicine
-        fullMedicine.reminders = arrayListOf()
+        fullMedicine.reminders = mutableListOf()
         medicine.unit = "pills"
         assertEquals(
             "test",
@@ -112,7 +112,7 @@ class MedicineHelperTest {
         val reminder = TestHelper.buildReminder(1, 1, "", 480, 1)
         reminder.outOfStockThreshold = 15.0
         reminder.outOfStockReminderType = Reminder.OutOfStockReminderType.ONCE
-        fullMedicine.reminders = arrayListOf(reminder)
+        fullMedicine.reminders = mutableListOf(reminder)
         assertEquals(
             "test (12 pills left ⚠)",
             MedicineHelper.getMedicineNameWithStockText(contextMock, fullMedicine).toString()
@@ -142,7 +142,7 @@ class MedicineHelperTest {
         val reminder = TestHelper.buildReminder(1, 1, "", 480, 1)
         reminder.outOfStockThreshold = 15.0
         reminder.outOfStockReminderType = Reminder.OutOfStockReminderType.ONCE
-        fullMedicine.reminders = arrayListOf(reminder)
+        fullMedicine.reminders = mutableListOf(reminder)
         medicine.amount = 0.0
         assertEquals(
             "t***",
--- a/app/src/test/java/com/futsch1/medtimer/schedulertests/ReminderSchedulerExpirationTest.kt
+++ b/app/src/test/java/com/futsch1/medtimer/schedulertests/ReminderSchedulerExpirationTest.kt
@@ -22,10 +22,10 @@ class ReminderSchedulerExpirationTest {
 
         medicine.reminders.add(reminder)
 
-        val medicineList: MutableList<FullMedicine> = ArrayList()
+        val medicineList: MutableList<FullMedicine> = mutableListOf()
         medicineList.add(medicine)
 
-        val reminderEventList: MutableList<ReminderEvent> = ArrayList()
+        val reminderEventList: MutableList<ReminderEvent> = mutableListOf()
 
         val scheduledReminders = scheduler.schedule(medicineList, reminderEventList)
         Assertions.assertTrue(scheduledReminders.isEmpty())
@@ -98,10 +98,10 @@ class ReminderSchedulerExpirationTest {
 
         medicine.reminders.add(reminder)
 
-        val medicineList: MutableList<FullMedicine> = ArrayList()
+        val medicineList: MutableList<FullMedicine> = mutableListOf()
         medicineList.add(medicine)
 
-        val reminderEventList: MutableList<ReminderEvent> = ArrayList()
+        val reminderEventList: MutableList<ReminderEvent> = mutableListOf()
 
         var scheduledReminders = scheduler.schedule(medicineList, reminderEventList)
         assertReminded(
--- a/app/src/test/java/com/futsch1/medtimer/schedulertests/ReminderSchedulerIntervalUnitTest.kt
+++ b/app/src/test/java/com/futsch1/medtimer/schedulertests/ReminderSchedulerIntervalUnitTest.kt
@@ -19,10 +19,10 @@ class ReminderSchedulerIntervalUnitTest {
         reminder.intervalStartsFromProcessed = false
         medicine.reminders.add(reminder)
 
-        val medicineList: MutableList<FullMedicine> = ArrayList()
+        val medicineList: MutableList<FullMedicine> = mutableListOf()
         medicineList.add(medicine)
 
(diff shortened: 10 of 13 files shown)


Futsch1:
@ThomasKiljanczykDev I am happy with this PR for now and I will merge it to the intermediate branch kotlin_conversion (feel free to review though). In this branch we can collect all changes until they are ready to be merged to main.

ThomasKiljanczykDev:
@Futsch1 I think you should just do a merge to main.
There's no benefit in collecting these changes in one big pull request.
It will be too big to review.

Lets be thorough with each individual PR and close this part of Kotlin migration.
Not merging to main will result in either recurring merge conflicts due to main updates, or total code freeze on main until you merge the big Kotlin pull request.

I'm still in the process of reviewing this pull request

EDIT: don't close this PR, we'll loose all the review comments. Just change the target

app/src/main/java/com/futsch1/medtimer/database/FullMedicine.kt
```diff
@@ -0,0 +1,43 @@
...
+    @Relation(parentColumn = "medicineId", entityColumn = "tagId", associateBy = Junction(MedicineToTag::class))
+    @Expose
+    var tags: List<Tag> = listOf()
+
+    @Relation(parentColumn = "medicineId", entityColumn = "medicineRelId")
+    @Expose
+    var reminders: MutableList<Reminder> = mutableListOf()
+
+    override fun equals(other: Any?): Boolean {
+        if (other == null || javaClass != other.javaClass) return false
```
ThomasKiljanczykDev: `if (other !is FullMedicine) return false` will eliminate null and automatically cast `other` to `FullMedicine` in lines following the return.

No need to creating the `that` variable.

app/src/main/java/com/futsch1/medtimer/database/FullMedicine.kt
```diff
@@ -0,0 +1,43 @@
...
+        val that = other as FullMedicine
+        return medicine == that.medicine && reminders == that.reminders && tags == that.tags
+    }
+
+    override fun hashCode(): Int {
```
ThomasKiljanczykDev, line 27: This hashCode implemnetation is incorrect.

Simple addition of hash codes does not eleminate the risk of collisions.
It's possible that different combinations of property hashcodes generate the same sum.

Use `Objects.hash()` to resolve this issue.

app/src/main/java/com/futsch1/medtimer/database/FullMedicine.kt
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    val isOutOfStock: Boolean
+        get() = this.isStockManagementActive && reminders.any { reminder: Reminder -> reminder.reminderType == Reminder.ReminderType.OUT_OF_STOCK && reminder.outOfStockThreshold >= medicine.amount }
+
+    val isStockManagementActive: Boolean
+        get() = (medicine.amount != 0.0 || hasStockReminder())
+
+    private fun hasStockReminder(): Boolean {
+        return reminders.any { reminder: Reminder? -> reminder!!.reminderType == Reminder.ReminderType.OUT_OF_STOCK }
```
ThomasKiljanczykDev: Do NOT suppress nulls unless you're 100% sure the value is not null.
You're generating code that is no different from the awful null handling of vanilla Java.
Here a simple null propagation `?` would be enough.

Also, no need for explicit `Reminder?` type specification.
Let the compiler do its job and infer it instead.
Here a null is not possible based on the reminders type definition.

app/src/main/java/com/futsch1/medtimer/database/JSONMedicineBackup.kt
```diff
@@ -8,7 +8,7 @@ class JSONMedicineBackup : JSONBackup<FullMedicine>(FullMedicine::class.java) {
     override fun createBackup(databaseVersion: Int, list: List<FullMedicine>): JsonElement {
         // Fix the medicines where the instructions are null
```
ThomasKiljanczykDev: Add `TODO` to the comment so it can be viewed in the in-IDE todo panel

app/src/main/java/com/futsch1/medtimer/database/JSONMedicineBackup.kt
```diff
@@ -8,7 +8,7 @@ class JSONMedicineBackup : JSONBackup<FullMedicine>(FullMedicine::class.java) {
...
         for (fullMedicine in list) {
-            fullMedicine.reminders.stream().filter { reminder: Reminder? -> reminder!!.instructions == null }
+            fullMedicine.reminders.filter { reminder: Reminder? -> reminder!!.instructions == null }
```
ThomasKiljanczykDev: No need for explicit `Reminder?`.
Based on the type definition, this collection cannot contain nulls.

app/src/main/java/com/futsch1/medtimer/database/JSONMedicineBackup.kt
```diff
@@ -8,7 +8,7 @@ class JSONMedicineBackup : JSONBackup<FullMedicine>(FullMedicine::class.java) {
...
                 .forEach { reminder: Reminder? -> reminder!!.instructions = "" }
```
ThomasKiljanczykDev: No need for explicit `Reminder?`.
Based on the type definition, this collection cannot contain nulls.

app/src/main/java/com/futsch1/medtimer/database/Medicine.kt
```diff
@@ -0,0 +1,104 @@
...
+    constructor(name: String) : this() {
+        this.name = name
+    }
+
+    fun hasExpired(): Boolean {
+        return expirationDate != 0L && expirationDate < LocalDate.now().toEpochDay()
+    }
+
+    override fun equals(other: Any?): Boolean {
+        if (other == null || javaClass != other.javaClass) return false
```
ThomasKiljanczykDev: Simple `other !is Medicine` would be enough for this condition.
No need for `javaClass` access.

This also eliminates for later need of `as Medicine`

app/src/main/java/com/futsch1/medtimer/database/Medicine.kt
```diff
@@ -0,0 +1,104 @@
...
+        return membersEqual(other as Medicine)
+    }
+
+    override fun hashCode(): Int {
+        return Objects.hash(medicineId, name, useColor, color, notificationImportance, iconId, amount, refillSizes, unit, notes, showNotificationAsAlarm)
+    }
+
+    private fun membersEqual(that: Medicine): Boolean {
```
ThomasKiljanczykDev: `that` is really vague, it does not mean anything. You should name it `other` to keep the same convention as for equals.

app/src/main/java/com/futsch1/medtimer/database/MedicineDao.kt
```diff
@@ -0,0 +1,162 @@
...
+    fun deleteMedicineToTags()
+
+    @get:Query("SELECT * FROM MedicineToTag")
+    val liveMedicineToTags: LiveData<List<MedicineToTag>>
+
+    @Query("SELECT COUNT(*) FROM Tag")
+    fun countTags(): Int
+
+    @get:Query("SELECT MAX(sortOrder) FROM Medicine")
+    val highestMedicineSortOrder: Double
```
ThomasKiljanczykDev, line 140: Can return null if the table is empty.
You can either change this to `Double?` or use `COALESCE` in the query to fallback to `0` on null.

app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
```diff
@@ -9,9 +9,9 @@ import java.util.LinkedList
 import java.util.concurrent.Callable
 import java.util.concurrent.ExecutionException
 
-open class MedicineRepository(val application: Application?) {
+open class MedicineRepository(val context: Context) {
```
ThomasKiljanczykDev, line 13: You should make the `context` a parameter instead of a property.

I see that `ReminderNotificationProcessor.buildReminderEvent` extracts the context from the repository.
This is a very convoluted dependency graph which has high risk of generating a memory risk.
Storing context as a property adds a new reference to the reference counter and might prevent the app from cleaning up the context depending what context is passed to this constructor.

Dependency injection using Hilt would make it way easier to have cleaner dependency graphs and prevent such mental gymnastics.

For now you could either leave a TODO comment or try tidy-up the dependencies a bit to don't have to expose the context via the repository.

app/src/main/java/com/futsch1/medtimer/database/Tag.kt
```diff
@@ -0,0 +1,30 @@
...
+class Tag(@field:Expose var name: String) {
+    @PrimaryKey(autoGenerate = true)
+    var tagId: Int = 0
+
+    @Ignore
+    constructor() : this("")
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || javaClass != other.javaClass) return false
```
ThomasKiljanczykDev, line 19: Could use `!is Tag` to do the implicit casting to `Tag` and remove the need of later `as Tag`

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -82,7 +81,7 @@ fun getIntervalTypeSummary(reminder: Reminder, context: Context): String {
 }
 
 fun linkedReminderString(reminder: Reminder, context: Context): String {
-    val medicineRepository = MedicineRepository(context.applicationContext as Application?)
+    val medicineRepository = MedicineRepository(context)
```
ThomasKiljanczykDev, line 84: Since you're building repositorie here, it means this is some service/manager and there should be a class which stores the repository as property and has all the methods from this 'helper'

app/src/main/java/com/futsch1/medtimer/medicine/EditMedicineFragment.kt
```diff
@@ -49,7 +49,7 @@ class EditMedicineFragment :
     private var color = 0
     private var notificationImportance: Spinner? = null
     private var selectIconButton: MaterialButton? = null
-    var notes: String? = null
+    var notes: String? = ""
```
ThomasKiljanczykDev, line 52: Should it really be nullable?
You could fallback or nullable assignments to `""` if the nullability is not desired here.

app/src/main/java/com/futsch1/medtimer/medicine/ReminderViewHolder.kt
ThomasKiljanczykDev, line 33: This is redundant. Just inline the `Dispatchers.IO` where it's needed within the viewholder.

app/src/main/java/com/futsch1/medtimer/medicine/ReminderViewHolder.kt
```diff
@@ -80,7 +80,7 @@ class ReminderViewHolder private constructor(
     private fun setupLongPress() {
         itemView.setOnCreateContextMenuListener { menu, _, _ ->
             menu.add(R.string.delete).setOnMenuItemClickListener {
-                LinkedReminderHandling(reminder, MedicineRepository(fragmentActivity.application), fragmentActivity.lifecycleScope).deleteReminder(
+                LinkedReminderHandling(reminder, MedicineRepository(fragmentActivity), fragmentActivity.lifecycleScope).deleteReminder(
```
ThomasKiljanczykDev, line 83: The repository instance should be stored viewholder wise. You should not spawn random repositories inline like this.

Inline services are not re-usable + you will quickly loose track of what is created where and what it references that it holds, i.e. you're inviting memory leaks.

app/src/main/java/com/futsch1/medtimer/overview/actions/ActionsFactory.kt
```diff
@@ -10,7 +10,7 @@ import com.futsch1.medtimer.overview.OverviewScheduledReminderEvent
 import kotlinx.coroutines.CoroutineScope
 
 fun createActions(event: OverviewEvent, fragmentActivity: FragmentActivity): Actions? {
-    val medicineRepository = MedicineRepository(fragmentActivity.application)
+    val medicineRepository = MedicineRepository(fragmentActivity)
```
ThomasKiljanczykDev, line 13: Since you're spawning the repository inside this method, this indicates the ActionsFactory should be class which stores the repository and exposes the repository dependent methods.

app/src/main/java/com/futsch1/medtimer/overview/EditEventSheetDialog.kt
```diff
@@ -80,7 +80,7 @@ class EditEventSheetDialog(val activity: FragmentActivity, val reminderEvent: Re
 
         activity.lifecycleScope.launch {
             withContext(ioDispatcher) {
```
ThomasKiljanczykDev, line 82: You could launch directly with the Dispatchers.IO instead of switching context right after starting the coroutine.

Also you should inline the `Dispatchers.IO`. Having a variable to hold it just convolutes the code.

app/src/main/java/com/futsch1/medtimer/overview/ShowMedicineSheetDialog.kt
```diff
@@ -26,7 +26,7 @@ class ShowMedicineSheetDialog(
 ) {
     init {
         activity.lifecycleScope.launch(ioDispatcher) {
```
ThomasKiljanczykDev, line 28: You should inline the `Dispatchers.IO` here

app/src/main/java/com/futsch1/medtimer/overview/ShowMedicineSheetDialog.kt
```diff
@@ -26,7 +26,7 @@ class ShowMedicineSheetDialog(
 ) {
     init {
```
ThomasKiljanczykDev, line 27: Instead of having a class with a magic `init`, add a public method which does what the init does.

Make it `suspend` and launch the coroutine where the method is called.

Righ now you're putting business logic into a wrong place. `init` is meant for additional initilzation. In most cases it should do non-async work only.

app/src/main/java/com/futsch1/medtimer/overview/ShowMedicineSheetDialog.kt
ThomasKiljanczykDev, line 33: Inline the dispatcher here, no need for storing it in a property

ThomasKiljanczykDev, line 22: A memory leak risk.
To change it to param right now.

You should mark it with a TODO to indicate there's a memory leak risk the activity property should migrated to a param.

Also consider you could pass context here instead.

app/src/main/java/com/futsch1/medtimer/reminders/ReminderContext.kt
```diff
@@ -22,7 +21,7 @@ import java.time.LocalDate
 import java.time.ZoneId
 
 class ReminderContext(private val context: Context) {
```
ThomasKiljanczykDev, line 23: A memory leak risk here.
You should mark it with a TODO note to modify the code to have `context` as param instead of field.

app/src/main/java/com/futsch1/medtimer/reminders/ReminderNotificationProcessor.kt
```diff
@@ -130,7 +130,7 @@ class ReminderNotificationProcessor(
 
                 Reminder.ReminderType.EXPIRATION_DATE -> {
                     reminderEvent.amount =
-                        TimeHelper.daysSinceEpochToDateString(medicineRepository.application!!.applicationContext, medicine.medicine.expirationDate)
+                        TimeHelper.daysSinceEpochToDateString(medicineRepository.context, medicine.medicine.expirationDate)
```
ThomasKiljanczykDev, line 133: Context should not be exposed via some intermediate services. If you need android context, pass it directly to the method.

In a proper app architecture, you should rarely see the context show up.
It has high potential for memory leaks.

app/src/main/java/com/futsch1/medtimer/widgets/WidgetLineProvider.kt
```diff
@@ -32,7 +32,7 @@ fun interface WidgetLineProvider {
 class NextRemindersLineProvider(val context: Context) : WidgetLineProvider {
     lateinit var scheduledReminders: List<ScheduledReminder>
     private val job: Job = CoroutineScope(SupervisorJob()).launch {
-        val medicineRepository = MedicineRepository(context.applicationContext as Application?)
+        val medicineRepository = MedicineRepository(context.applicationContext as Application)
```
ThomasKiljanczykDev, line 35: No need for the `Application` cast here

Futsch1:
I see the following benefit: I will be able to still work on features in parallel (especially in areas that are not touched by the Kotlin migration) and will also be able to deliver bugfixes from a stable basis. Once we start merging the refactoring changes, the risk of breaking something is bigger and I would like to take some time to test a beta version of this before releasing the next version.

I consider `main` my stable branch and I would like to use the `kotlin_conversion` branch as the one to integrate all the migration changes to.

So I prefer to keep it separated - also I do not know how fast we will be in collaboration and how long this refactoring will take. As of now, `kotlin_conversion` and `main` have not deviated, and as long as it stays that way, we will have no issue anyway.

app/src/main/java/com/futsch1/medtimer/GenerateTestData.kt
```diff
@@ -111,7 +111,7 @@ class GenerateTestData(private val viewModel: MedicineViewModel, val withEvents:
             if (stock > 0) {
                 medicine.amount = stock
                 medicine.unit = "pills"
-                medicine.refillSizes.add(20.0)
+                medicine.refillSizes = arrayListOf(20.0)
```
ThomasKiljanczykDev, line 114: Is the `arrayListOf` really necessary here.
It just adds noise with no benefits.
99% of cases no-one cares about the actual list implementation (here you specify you want an ArrayList specifically).

app/src/test/java/com/futsch1/medtimer/database/MigrationTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    @BeforeEach
+    fun setUp() {
+        helper = MigrationTestHelper(
+            InstrumentationRegistry.getInstrumentation(),
+            MedicineRoomDatabase::class.java
+        )
+    }
+
+    @Test
+    @Throws(IOException::class)
```
ThomasKiljanczykDev, line 31: Does anyone care what a test throws?
It's probably not needed here.
If a test throws `IOException`, you'll see that as a test failed.

app/src/test/java/com/futsch1/medtimer/database/MigrationTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    fun testMigrate22To23() {
+        // Create version 22 of the database.
+        var db = helper.createDatabase(testDb, 22)
```
ThomasKiljanczykDev, line 34: A good practice is to split tests using the AAA pattern (Arrange Act Assert).

I usually add comments `// Arrange`, `// Act` and `// Assert` to indicate the specific parts of the test.

Here arrange is the creation of the db and insert data to it, act is runMigrationsAndValidate, and assert is the rest of the test

ThomasKiljanczykDev, line 34: You should probably do `helper.createDatabase(testDb, 22).use { db -> ... }` instead to make sure the db is closed no matter what happens in the test.

app/src/test/java/com/futsch1/medtimer/database/MigrationTest.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        // Insert some data with NULL values
+        db.execSQL("INSERT INTO Medicine (medicineName, color, useColor, notificationImportance, iconId, amount, refillSizes, unit, sortOrder, notes, showNotificationAsAlarm, productionDate, expirationDate) VALUES (NULL, 0, 0, 3, 0, 0.0, NULL, NULL, 1.0, '', 0, 0, 0)")
+        db.execSQL("INSERT INTO Tag (name, tagId) VALUES (NULL, 1)")
+
+        // Prepare for the next version.
+        db.close()
+
+        // Re-open the database with version 23 and validate
+        db = helper.runMigrationsAndValidate(testDb, 23, true, MedicineRoomDatabase.MIGRATION_22_23)
```
ThomasKiljanczykDev, line 44: You should not re-use variables like that, it complicates the way the code should be understood.
This is a different instance of the db than the one used in the arrange part.

Also you could apply the `.use {db -> ... }`  as advised in the previous comment.

app/src/test/java/com/futsch1/medtimer/database/MigrationTest.kt
```diff
@@ -0,0 +1,61 @@
...
+
+        // Verify data was migrated properly
+        var cursor = db.query("SELECT medicineName, refillSizes, unit FROM Medicine")
```
ThomasKiljanczykDev, line 47: You could use `.use {}` yo make sure the cursor is always closed at the end.

app/src/test/java/com/futsch1/medtimer/database/MigrationTest.kt
```diff
@@ -0,0 +1,61 @@
...
+        cursor.moveToFirst()
+        assertEquals("", cursor.getString(0))
+        assertEquals("[]", cursor.getString(1))
+        assertEquals("", cursor.getString(2))
+        cursor.close()
+
+        cursor = db.query("SELECT name FROM Tag")
```
ThomasKiljanczykDev, line 54: You could use `.use {}` to make sure the cursor is always closed at the end.

app/src/test/java/com/futsch1/medtimer/schedulertests/ReminderSchedulerActiveDayOfMonthUnitTest.java
```diff
@@ -24,8 +24,8 @@ void scheduleDayOfMonth() {
 
         FullMedicine medicineWithReminders = TestHelper.buildFullMedicine(1, "Test");
         Reminder reminder = TestHelper.buildReminder(1, 1, "1", 480, 1);
-        reminder.activeDaysOfMonth = 0x7;
-        medicineWithReminders.reminders.add(reminder);
+        reminder.setActiveDaysOfMonth(0x7);
```
ThomasKiljanczykDev, line 27: Why hex here? Simple `7` would do the same thing

app/src/test/java/com/futsch1/medtimer/MedicineHelperTest.kt
```diff
@@ -86,7 +86,7 @@ class MedicineHelperTest {
         val medicine = Medicine("test")
         val fullMedicine = FullMedicine()
         fullMedicine.medicine = medicine
-        fullMedicine.reminders = listOf()
+        fullMedicine.reminders = arrayListOf()
```
ThomasKiljanczykDev: No need for the explicit list implementation, could use `listOf()` to reduce the noise

app/src/test/java/com/futsch1/medtimer/MedicineHelperTest.kt
```diff
@@ -112,7 +112,7 @@ class MedicineHelperTest {
         val reminder = TestHelper.buildReminder(1, 1, "", 480, 1)
         reminder.outOfStockThreshold = 15.0
         reminder.outOfStockReminderType = Reminder.OutOfStockReminderType.ONCE
-        fullMedicine.reminders = listOf(reminder)
+        fullMedicine.reminders = arrayListOf(reminder)
```
ThomasKiljanczykDev: No need for the explicit list implementation, could use `listOf()` to reduce the noise

app/src/test/java/com/futsch1/medtimer/MedicineHelperTest.kt
```diff
@@ -142,7 +142,7 @@ class MedicineHelperTest {
...
```
ThomasKiljanczykDev: No need for the explicit list implementation, could use `listOf()` to reduce the noise

app/build.gradle.kts
```diff
@@ -22,7 +23,7 @@ android {
         multiDexEnabled = true
         targetSdk = 36
         versionCode = 161
-        versionName = "1.22.9"
+        versionName = "1.23.0"
```
ThomasKiljanczykDev, line 26: Is it really necessary to bump the version just yet?
Are you planning a release specifically after this PR is merged?

app/src/main/java/com/futsch1/medtimer/database/JSONMedicineBackup.kt
Futsch1: This comment might be formulated incorrectly - I had in the past a crash that instructions were null in the database, which I will fix here. It is no TODO comment

app/src/main/java/com/futsch1/medtimer/database/MedicineDao.kt
Futsch1: Nice find!

app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
Futsch1: Yes, that is indeed a pain. I used the pattern of Service Locator for the reminder handling part, but there is this one location where it broke. I did not have the time to refactor it back then and I would like to keep the Service Locator pattern in general, but we should clean it up. Going with a TODO would be fine for me now.

app/src/test/java/com/futsch1/medtimer/schedulertests/ReminderSchedulerActiveDayOfMonthUnitTest.java
Futsch1: It's a bit mask, that's what this should have indicated (so if you change it later, this could become more obvious). But it might also show my embedded software background ;-)

Let's keep it for now, ok?

app/build.gradle.kts
Futsch1: It is not necessary, but I did it since I changed the database layout. Database version number corresponds with the minor number and I did it to not forget to set it correspondingly. Especially the version code will change, since I plan to do a release before this is merged to main. I kinda do my version bumps a bit uncoordinated actually and it does not correspond to a desire to immediately release.

ThomasKiljanczykDev:
@Futsch1 the usual rule in code reviews and comments is that the author of the PR addresses the comment or starts a discussion.

If a comment is addressed the author should reply that it's done, or shortly describe how it's done if there's some deviation.
Resolving comments by the author makes it harder for the reviewer to track what has been done, what has been skipped and why, what will be done in later tasks.

Without knowing these things I might make the same review comments even when I could skip some of them knowing there's a higher level plan to address some of the code quality issues.

Futsch1:
> @Futsch1 the usual rule in code reviews and comments is that the author of the PR addresses the comment or starts a discussion.
> 
> If a comment is addressed the author should reply that it's done, or shortly describe how it's done if there's some deviation. Resolving comments by the author makes it harder for the reviewer to track what has been done, what has been skipped and why, what will be done in later tasks.
> 
> Without knowing these things I might make the same review comments even when I could skip some of them knowing there's a higher level plan to address some of the code quality issues.

Understood - I had it rather the way that I know what I addressed and if I finished working on all the comments, but I will keep the comments open then in the future.

Edit: I think both approaches have their pros and cons (and I am not sure if there's really a general rule), but for the time being, let's make the PR reviewer resolve as you proposed.

app/src/main/java/com/futsch1/medtimer/database/Medicine.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    override fun hashCode(): Int {
+        return Objects.hash(medicineId, name, useColor, color, notificationImportance, iconId, amount, refillSizes, unit, notes, showNotificationAsAlarm)
+    }
+
+    private fun membersEqual(other: Medicine): Boolean {
+        return medicineId == other.medicineId &&
+                name == other.name && useColor == other.useColor && color == other.color && notificationImportance == other.notificationImportance && iconId == other.iconId && amount == other.amount &&
+                refillSizes == other.refillSizes &&
+                unit == other.unit &&
```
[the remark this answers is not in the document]
ThomasKiljanczykDev: That's true, but there's no other way for now.

Once the DB entities are abstracted via domain models, this `equals` implementation can be moved to the domain model, or the domain model can be defined as a `data class` and the custom equals implementation will be redundant.

app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
```diff
@@ -9,9 +9,10 @@ import java.util.LinkedList
 import java.util.concurrent.Callable
 import java.util.concurrent.ExecutionException
 
-open class MedicineRepository(val application: Application?) {
+// TODO: Context should not be a property, but ReminderNotificationProcessor.buildReminderEvent must be changed before this is possible
```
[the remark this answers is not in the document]
ThomasKiljanczykDev: I normal circumstances that would be true. That being said, with a refactor of this size, it's not possible to not leave unresolved TODO comments for now.

app/src/main/java/com/futsch1/medtimer/database/Reminder.kt
```diff
@@ -207,12 +183,12 @@ class Reminder(@JvmField var medicineRelId: Int) {
...
-                amount == that.amount &&
-                days == that.days && active == that.active && periodStart == that.periodStart && periodEnd == that.periodEnd && activeDaysOfMonth == that.activeDaysOfMonth && linkedReminderId == that.linkedReminderId && intervalStart == that.intervalStart && intervalStartsFromProcessed == that.intervalStartsFromProcessed && variableAmount == that.variableAmount && intervalStartTimeOfDay == that.intervalStartTimeOfDay && intervalEndTimeOfDay == that.intervalEndTimeOfDay && windowedInterval == that.windowedInterval &&
-                outOfStockThreshold == that.outOfStockThreshold && outOfStockReminderType == that.outOfStockReminderType && expirationReminderType == that.expirationReminderType
+    private fun membersEqual(other: Reminder): Boolean {
+        return reminderId == other.reminderId && medicineRelId == other.medicineRelId && timeInMinutes == other.timeInMinutes && createdTimestamp == other.createdTimestamp && consecutiveDays == other.consecutiveDays && pauseDays == other.pauseDays &&
+                instructions == other.instructions && cycleStartDay == other.cycleStartDay &&
+                amount == other.amount &&
```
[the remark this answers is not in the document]
ThomasKiljanczykDev: Not possible to resolve for now. Needs domain model abstraction first so `data class` can be used

app/src/main/java/com/futsch1/medtimer/database/ReminderEvent.kt
```diff
@@ -126,12 +108,12 @@ class ReminderEvent {
...
-                amount == o.amount && color == o.color && useColor == o.useColor && status == o.status && remindedTimestamp == o.remindedTimestamp && processedTimestamp == o.processedTimestamp && reminderId == o.reminderId && notificationId == o.notificationId && iconId == o.iconId && remainingRepeats == o.remainingRepeats && stockHandled == o.stockHandled && askForAmount == o.askForAmount &&
-                tags == o.tags && lastIntervalReminderTimeInMinutes == o.lastIntervalReminderTimeInMinutes &&
-                notes == o.notes && reminderType == o.reminderType
+    private fun membersEqual(other: ReminderEvent): Boolean {
+        return reminderEventId == other.reminderEventId &&
+                medicineName == other.medicineName &&
+                amount == other.amount && color == other.color && useColor == other.useColor && status == other.status && remindedTimestamp == other.remindedTimestamp && processedTimestamp == other.processedTimestamp && reminderId == other.reminderId && notificationId == other.notificationId && iconId == other.iconId && remainingRepeats == other.remainingRepeats && stockHandled == other.stockHandled && askForAmount == other.askForAmount &&
```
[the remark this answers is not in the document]
ThomasKiljanczykDev: Not possible to resolve for now. Needs domain model abstraction first so data class can be used

app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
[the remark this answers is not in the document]
ThomasKiljanczykDev: @Futsch1 can this rule be suppressed? It does not make sense in the context of the Kotlin migration and the overall architectural overhaul.

We need to have a way to track things to refactor and TODO comments are the easiest way

PR #1265, opened 2026-03-07
Convert helpers to kotlin

Futsch1:
@ThomasKiljanczykDev This was the second PR I prepared. I will stop here in the conversion and leave the rest to you if you agree. I would in the meantime focus on release v1.22.9.

Sorry for the back and forth regarding the collaborators, I would like to wait a bit more until I give this right to you. Still greatly appreciate your feedback and the help you're providing!

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
+++ b/app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
@@ -29,20 +29,21 @@ object FileHelper {
         return true
     }
 
-    fun readFromUri(uri: Uri?, resolver: ContentResolver): String? {
-        if (uri != null && uri.path != null) {
-            try {
-                resolver.openInputStream(uri).use { inputStream ->
-                    InputStreamReader(inputStream, StandardCharsets.UTF_8).use { inputStreamReader ->
-                        val stringBuilder = getStringBuilder(inputStreamReader)
-                        return stringBuilder.toString()
-                    }
-                }
-            } catch (e: IOException) {
-                Log.e(LogTags.BACKUP, e.toString())
-            }
+    fun readFromUri(uri: Uri, resolver: ContentResolver): String? {
+        if (uri.path == null) {
+            return null
         }
-        return null
+        try {
+            resolver.openInputStream(uri).use { inputStream ->
+                InputStreamReader(inputStream, StandardCharsets.UTF_8).use { inputStreamReader ->
+                    val stringBuilder = getStringBuilder(inputStreamReader)
+                    return stringBuilder.toString()
+                }
+            }
+        } catch (e: IOException) {
+            Log.e(LogTags.BACKUP, e.toString())
+            return null
+        }
     }
 
     @Throws(IOException::class)
--- a/app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
+++ b/app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
@@ -102,18 +102,19 @@ private fun timeBasedReminderString(
     strings: MutableList<String>,
     context: Context
 ) {
-    val weekdayLimited =
-        !reminder.days.all { day: Boolean -> day }
-    val dayOfMonthLimited = (reminder.activeDaysOfMonth and 0x7FFFFFFF) != 0x7FFFFFFF
-    val never = reminder.days.none { day: Boolean -> day } || (reminder.activeDaysOfMonth and 0x7FFFFFFF) == 0
+    val never = reminder.days.none { it } || (reminder.activeDaysOfMonth and 0x7FFFFFFF) == 0
     if (never) {
         strings.add(context.getString(R.string.never))
     } else {
+        val dayOfMonthLimited = (reminder.activeDaysOfMonth and 0x7FFFFFFF) != 0x7FFFFFFF
+        val hasWeekdayRestriction =
+            reminder.days.any { !it }
+
         buildReminderStrings(
             context,
             strings,
             reminder,
-            ReminderProperties(weekdayLimited, dayOfMonthLimited)
+            ReminderProperties(hasWeekdayRestriction, dayOfMonthLimited)
         )
     }
 }
@@ -143,21 +144,13 @@ private fun buildReminderStrings(
 
 fun remindersSummary(reminders: List<Reminder>, context: Context): String {
     val reminderTimes = timeBasedRemindersSummary(
-        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.TIME_BASED },
+        reminders.filter { r -> r.reminderType == Reminder.ReminderType.TIME_BASED },
         context
-    ) + getRemindersSummary(
-        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.LINKED },
-        ({ r: Reminder -> linkedReminderSummaryString(r, context) })
-    ) + getRemindersSummary(
-        reminders.filter { r: Reminder -> r.isInterval },
-        ({ r: Reminder -> intervalBasedReminderString(r, context) })
-    ) + getRemindersSummary(
-        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.OUT_OF_STOCK },
-        ({ r: Reminder -> outOfStockReminderString(r, context) })
-    ) + getRemindersSummary(
-        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.EXPIRATION_DATE },
-        ({ _: Reminder -> context.getString(R.string.expiration_date) })
-    )
+    ) +
+            reminders.filter { r -> r.reminderType == Reminder.ReminderType.LINKED }.map { linkedReminderSummaryString(it, context) } +
+            reminders.filter { r -> r.isInterval }.map { intervalBasedReminderString(it, context) } +
+            reminders.filter { r -> r.reminderType == Reminder.ReminderType.OUT_OF_STOCK }.map { outOfStockReminderString(it, context) } +
+            reminders.filter { r -> r.reminderType == Reminder.ReminderType.EXPIRATION_DATE }.map { context.getString(R.string.expiration_date) }
 
     val len = reminderTimes.size
     return context.resources.getQuantityString(
@@ -166,13 +159,6 @@ fun remindersSummary(reminders: List<Reminder>, context: Context): String {
         len,
         java.lang.String.join("; ", reminderTimes)
     )
-}
-
-fun getRemindersSummary(
-    reminders: List<Reminder>,
-    toStringFunction: (Reminder) -> String
-): List<String> {
-    return reminders.map { r: Reminder -> toStringFunction(r) }
 }
 
 fun linkedReminderSummaryString(reminder: Reminder, context: Context): String {
@@ -199,13 +185,7 @@ private fun timeBasedRemindersSummary(
     reminders: List<Reminder>,
     context: Context
 ): List<String> {
-    val reminderTimes: MutableList<String> = mutableListOf()
-    val timesInMinutes =
-        reminders.map { r: Reminder -> r.timeInMinutes }.sorted()
-    for (minute in timesInMinutes) {
-        reminderTimes.add(TimeHelper.minutesToTimeString(context, minute.toLong()))
-    }
-    return reminderTimes
+    return reminders.map { r -> r.timeInMinutes }.sorted().map { TimeHelper.minutesToTimeString(context, it.toLong()) }
 }
 
 fun getCyclicReminderString(reminder: Reminder, context: Context): String {
--- a/app/src/main/java/com/futsch1/medtimer/helpers/SwipeHelper.kt
+++ b/app/src/main/java/com/futsch1/medtimer/helpers/SwipeHelper.kt
@@ -38,7 +38,6 @@ abstract class SwipeHelper protected constructor(context: Context, dragDirs: Int
         actionState: Int, isCurrentlyActive: Boolean
     ) {
         val itemView = viewHolder.itemView
-        val itemHeight = itemView.bottom - itemView.top
         val isCanceled = (dX == 0f) && !isCurrentlyActive
 
         if (isCanceled) {
@@ -48,6 +47,8 @@ abstract class SwipeHelper protected constructor(context: Context, dragDirs: Int
         }
 
         if (dX < 0) {
+            val itemHeight = itemView.bottom - itemView.top
+            
             drawSwipeBar(c, dX, itemView, itemHeight)
         }
         super.onChildDraw(c, recyclerView, viewHolder, dX, dY, actionState, isCurrentlyActive)
--- a/app/src/main/java/com/futsch1/medtimer/medicine/MedicineViewAdapter.kt
+++ b/app/src/main/java/com/futsch1/medtimer/medicine/MedicineViewAdapter.kt
@@ -42,8 +42,9 @@ class MedicineViewAdapter(activity: FragmentActivity, medicineRepository: Medici
     }
 
     override fun onMoved(fromPosition: Int, toPosition: Int) {
-        val list: List<FullMedicine> = ArrayList(currentList)
         if (toPosition != -1) {
+            val list = currentList.toList()
+
             Collections.swap(list, fromPosition, toPosition)
             submitList(list)
         }
--- a/app/src/main/java/com/futsch1/medtimer/BackupManager.kt
+++ b/app/src/main/java/com/futsch1/medtimer/BackupManager.kt
@@ -34,7 +34,7 @@ class BackupManager(
     private val fragment: Fragment,
     private val menu: Menu,
     private val medicineViewModel: MedicineViewModel,
-    private val openFileLauncher: ActivityResultLauncher<Intent?>,
+    private val openFileLauncher: ActivityResultLauncher<Intent>,
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
     private val mainDispatcher: CoroutineDispatcher = Dispatchers.Main
 ) {
@@ -135,7 +135,7 @@ class BackupManager(
         }
     }
 
-    fun fileSelected(data: Uri?) {
+    fun fileSelected(data: Uri) {
         fragment.lifecycleScope.launch(ioDispatcher) {
             val json = FileHelper.readFromUri(data, context.contentResolver)
 
--- a/app/src/main/java/com/futsch1/medtimer/OptionsMenu.kt
+++ b/app/src/main/java/com/futsch1/medtimer/OptionsMenu.kt
@@ -53,22 +53,22 @@ class OptionsMenu(
     private val mainDispatcher: CoroutineDispatcher = Dispatchers.Main
 ) : EntityEditOptionsMenu {
     private val context: Context = fragment.requireContext()
-    private val openFileLauncher: ActivityResultLauncher<Intent?>
+    private val openFileLauncher: ActivityResultLauncher<Intent>
     private val idlingResource: SimpleIdlingResource = SimpleIdlingResource("OptionsMenu_" + fragment.javaClass.getName())
     private var menu: Menu? = null
     private var backupManager: BackupManager? = null
 
     init {
         this.openFileLauncher =
-            fragment.registerForActivityResult<Intent?, ActivityResult?>(StartActivityForResult()) { result: ActivityResult? ->
-                if (result!!.resultCode == Activity.RESULT_OK && result.data != null) {
-                    this.fileSelected(result.data!!.data)
+            fragment.registerForActivityResult<Intent, ActivityResult>(StartActivityForResult()) { result: ActivityResult ->
+                if (result.resultCode == Activity.RESULT_OK) {
+                    result.data?.data?.let { this.fileSelected(it) }
                 }
             }
         idlingResource.setIdle()
     }
 
-    fun fileSelected(data: Uri?) {
+    fun fileSelected(data: Uri) {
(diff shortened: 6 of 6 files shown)


app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class DeleteHelper(private val context: Context?) {
+    fun deleteItem(messageStringId: Int, yesClicked: ButtonCallback, noClicked: ButtonCallback) {
+        if (context == null) {
+            return
+        }
+        val builder = AlertDialog.Builder(context)
+        builder.setTitle(R.string.confirm)
+        builder.setMessage(messageStringId)
+        builder.setPositiveButton(R.string.yes) { _: DialogInterface?, _: Int -> yesClicked.onButtonClick() }
```
ThomasKiljanczykDev: Since you're just ignoring the lambda parameters, you can skip the types and just leave `_` in place

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
```diff
@@ -0,0 +1,25 @@
...
+        builder.setNegativeButton(R.string.cancel) { _: DialogInterface?, _: Int -> noClicked.onButtonClick() }
```
ThomasKiljanczykDev: Since you're just ignoring the lambda parameters, you can skip the types and just leave _ in place

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
```diff
@@ -0,0 +1,25 @@
...
+        builder.setOnCancelListener { _: DialogInterface? -> noClicked.onButtonClick() }
```
ThomasKiljanczykDev: Single parameter lambda with ignore parameter. Here you can remove the parameter and `->` part altogether

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
```diff
@@ -0,0 +1,25 @@
+package com.futsch1.medtimer.helpers
+
+import android.app.AlertDialog
+import android.content.Context
+import android.content.DialogInterface
+import com.futsch1.medtimer.R
+
+class DeleteHelper(private val context: Context?) {
```
ThomasKiljanczykDev: `context` most likely should not be nullable. It makes better sense to just not create a DeleteHelper if context is null for whatever reason.

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
```diff
@@ -0,0 +1,25 @@
...
+    fun deleteItem(messageStringId: Int, yesClicked: ButtonCallback, noClicked: ButtonCallback) {
```
ThomasKiljanczykDev: It would most likely be more sensible to pass the context here instead of holding it as a property in the helper class.

This would effectively convert the helper class into a helper `object`.

ThomasKiljanczykDev: Instead of having the `ButtonCallback`, it would be more Kotlin-like if you just defined the types as `() -> Unit`

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
ThomasKiljanczykDev: I've seen usages of the constructor where there is a type definition clash.
There were 2 cases, first in which the context came from a parameter that could be changed to be non-nullable, second where `fragment.context` was used.

`fragment.context` can be replaced with `fragment.requireContext()` to ensure the context is non-nullable.

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
```diff
@@ -0,0 +1,77 @@
...
+                writer.write(content)
+            }
+        } catch (e: IOException) {
+            Log.e(LogTags.BACKUP, e.toString())
+            return false
+        }
+        return true
+    }
+
+    fun readFromUri(uri: Uri?, resolver: ContentResolver): String? {
```
ThomasKiljanczykDev: Since this returns `null` when the `uri` is null, what's the point of having a nullable `uri` parameter?

Maybe it would make more sense to handle the null `uri` in the upstream usages and have cleaner code here.

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
```diff
@@ -0,0 +1,77 @@
...
+        if (uri != null && uri.path != null) {
```
ThomasKiljanczykDev: You could reduce the code depth by inverting this condition to return null immediately.

My opinion is that code should flow in a direction of success.
Any failure/safeguard path should resolve at the moment where it's checked.

If you were to invert the condition, another refactor would be to move the final return null into the catch block to clearly indicate the null is returned only if an error happens.

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
```diff
@@ -0,0 +1,77 @@
...
+            try {
+                resolver.openInputStream(uri).use { inputStream ->
+                    InputStreamReader(inputStream, StandardCharsets.UTF_8).use { inputStreamReader ->
+                        val stringBuilder = getStringBuilder(inputStreamReader)
+                        return stringBuilder.toString()
+                    }
+                }
+            } catch (e: IOException) {
+                Log.e(LogTags.BACKUP, e.toString())
```
ThomasKiljanczykDev: The standard for Android development is to co-locate TAG constant with the code that uses the Log.
I.e. keep it located either as a const in a companion object of a class, or as a const within a file in case the code is not a class.

A single `LogTags` object with grow indefinitely and has little point to exist.

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -105,10 +103,9 @@ private fun timeBasedReminderString(
     context: Context
 ) {
     val weekdayLimited =
-        !reminder.days.stream().allMatch { day: Boolean -> day }
+        !reminder.days.all { day: Boolean -> day }
```
ThomasKiljanczykDev: `val hasWeekdayRestriction = reminder.days.any { !it }` would be more idiomatic and probably easier to understand

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -105,10 +103,9 @@ private fun timeBasedReminderString(
     context: Context
 ) {
     val weekdayLimited =
```
ThomasKiljanczykDev: Pre-mature declaration. It's used only in the `else` branch, so could be move to there.

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -105,10 +103,9 @@ private fun timeBasedReminderString(
...
-        !reminder.days.stream().allMatch { day: Boolean -> day }
+        !reminder.days.all { day: Boolean -> day }
     val dayOfMonthLimited = (reminder.activeDaysOfMonth and 0x7FFFFFFF) != 0x7FFFFFFF
```
ThomasKiljanczykDev: Pre-mature declaration. It's used only in the `else` branch, so could be move to there.

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -105,10 +103,9 @@ private fun timeBasedReminderString(
...
-    val never = reminder.days.stream()
-        .noneMatch { day: Boolean -> day } || (reminder.activeDaysOfMonth and 0x7FFFFFFF) == 0
+    val never = reminder.days.none { day: Boolean -> day } || (reminder.activeDaysOfMonth and 0x7FFFFFFF) == 0
```
ThomasKiljanczykDev: No need to specify boolean for the day arg.
Could also be replace with just `it`.

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -146,23 +143,19 @@ private fun buildReminderStrings(
 
 fun remindersSummary(reminders: List<Reminder>, context: Context): String {
     val reminderTimes = timeBasedRemindersSummary(
-        reminders.stream()
-            .filter { r: Reminder -> r.reminderType == Reminder.ReminderType.TIME_BASED },
+        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.TIME_BASED },
```
ThomasKiljanczykDev: No need to specify the lambda parameter type explicitly

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -146,23 +143,19 @@ private fun buildReminderStrings(
...
         context
     ) + getRemindersSummary(
-        reminders.stream().filter { r: Reminder -> r.reminderType == Reminder.ReminderType.LINKED },
+        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.LINKED },
```
ThomasKiljanczykDev: No need to specify the lambda parameter type explicitly

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -146,23 +143,19 @@ private fun buildReminderStrings(
...
         ({ r: Reminder -> linkedReminderSummaryString(r, context) })
     ) + getRemindersSummary(
-        reminders.stream()
-            .filter { r: Reminder -> r.isInterval },
+        reminders.filter { r: Reminder -> r.isInterval },
```
ThomasKiljanczykDev: No need to specify the lambda parameter type explicitly

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -146,23 +143,19 @@ private fun buildReminderStrings(
...
         ({ r: Reminder -> intervalBasedReminderString(r, context) })
     ) + getRemindersSummary(
-        reminders.stream()
-            .filter { r: Reminder -> r.reminderType == Reminder.ReminderType.OUT_OF_STOCK },
+        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.OUT_OF_STOCK },
```
ThomasKiljanczykDev: No need to specify the lambda parameter type explicitly

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -146,23 +143,19 @@ private fun buildReminderStrings(
...
         ({ r: Reminder -> outOfStockReminderString(r, context) })
     ) + getRemindersSummary(
-        reminders.stream()
-            .filter { r: Reminder -> r.reminderType == Reminder.ReminderType.EXPIRATION_DATE },
+        reminders.filter { r: Reminder -> r.reminderType == Reminder.ReminderType.EXPIRATION_DATE },
```
ThomasKiljanczykDev: No need to specify the lambda parameter type explicitly

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -176,11 +169,10 @@ fun remindersSummary(reminders: List<Reminder>, context: Context): String {
 }
 
 fun getRemindersSummary(
```
ThomasKiljanczykDev: `getRemindersSummary` seems totally unneeded.
You could just add the `.map` where it's invoked.
In this form its just a thin-wrapper for `map`.

If the code where the actual mapping is done is too convoluted, you can always split it into variables and then concatenate the collections.

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
```diff
@@ -204,12 +196,12 @@ fun linkedReminderSummaryString(reminder: Reminder, context: Context): String {
...
     context: Context
-): ArrayList<String> {
-    val reminderTimes = ArrayList<String>()
+): List<String> {
+    val reminderTimes: MutableList<String> = mutableListOf()
     val timesInMinutes =
-        reminders.mapToInt { r: Reminder -> r.timeInMinutes }.sorted().toArray()
+        reminders.map { r: Reminder -> r.timeInMinutes }.sorted()
     for (minute in timesInMinutes) {
         reminderTimes.add(TimeHelper.minutesToTimeString(context, minute.toLong()))
```
ThomasKiljanczykDev: The `TimeHelper.minutesToTimeString` could be folded into the `reminders.map`.
If the sorted() can't be done on the TimeString, then you can always chain another `.map`.

In this case you could return `reminders.map { ... }` immediately.

app/src/main/java/com/futsch1/medtimer/helpers/SwipeHelper.kt
```diff
@@ -0,0 +1,148 @@
...
+        setDefaultSwipeDirs(SWIPE_DIRS)
+    }
+
+    override fun onChildDrawOver(
+        c: Canvas, recyclerView: RecyclerView,
+        viewHolder: RecyclerView.ViewHolder, dX: Float, dY: Float,
+        actionState: Int, isCurrentlyActive: Boolean
+    ) {
+        val itemView = viewHolder.itemView
+        val itemHeight = itemView.bottom - itemView.top
```
ThomasKiljanczykDev: This variable is only needed for the `dx < 0` branch.
Could be placed there.

app/src/main/java/com/futsch1/medtimer/medicine/MedicineViewAdapter.kt
```diff
@@ -42,7 +42,7 @@ class MedicineViewAdapter(activity: FragmentActivity, medicineRepository: Medici
     }
 
     override fun onMoved(fromPosition: Int, toPosition: Int) {
-        val list: MutableList<FullMedicine?> = ArrayList(currentList)
+        val list: List<FullMedicine> = ArrayList(currentList)
```
ThomasKiljanczykDev: could use `currentList.toList()` instead.
The explicit variable type won't be needed in that case.

Also could be moved into the `if` branch since it's only needed there

app/src/main/java/com/futsch1/medtimer/helpers/SwipeHelper.kt
```diff
@@ -0,0 +1,148 @@
...
+        intrinsicHeight = swipeIcon.intrinsicHeight
+        intrinsicWidth = swipeIcon.intrinsicWidth
+
+        setDefaultSwipeDirs(SWIPE_DIRS)
+    }
+
+    override fun onChildDrawOver(
+        c: Canvas, recyclerView: RecyclerView,
+        viewHolder: RecyclerView.ViewHolder, dX: Float, dY: Float,
+        actionState: Int, isCurrentlyActive: Boolean
```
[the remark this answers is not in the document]
ThomasKiljanczykDev: Invalid comment. We don't have any way to modify this signature since it comes outside of this codebase.

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
Futsch1: What I like to do is to group logs logically, not by class or object. So I want to log backup related stuff, so I can filter the logs for the backup tag when I debug a backup issue (same goes for reminders etc.). It often does not make sense to log only for a single class, because most issues that require extensive debugging cover multiple classes (single class issues should be reproducible/solvable via unit tests). That's why I also do not consider it to grow so much, since the potential topics (aka features) of the app won't grow so quickly.

Futsch1:
@ThomasKiljanczykDev Thanks for the review. I should be done addressing the feedback.

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
ThomasKiljanczykDev: https://developer.android.com/reference/android/util/Log

ThomasKiljanczykDev:
@Futsch1 some of my comments in the feed have been folded into the hidden conversations thingy. Make sure to unfold these and check them as well.

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
Futsch1: That's what they say... But I'm not convinced to change it, so let's leave it as it is for now.

Futsch1:
> @Futsch1 some of my comments in the feed have been folded into the hidden conversations thingy. Make sure to unfold these and check them as well.

@ThomasKiljanczykDev I think I got all of them, they are all marked as "Outdated", meaning I changed the corresponding line...

Edit: Also commented them as done now. Tbh, resolving them on my side felt much more straightforward...

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/DeleteHelper.kt
Futsch1: Done

Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/FileHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SummaryHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/SwipeHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/medicine/MedicineViewAdapter.kt
Futsch1: Done

PR #1266, opened 2026-03-07
1260 kotlin conversion statistics feature

ThomasKiljanczykDev:
@Futsch1 please review my changes

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/statistics/ChartsFragment.kt
+++ b/app/src/main/java/com/futsch1/medtimer/statistics/ChartsFragment.kt
@@ -10,12 +10,16 @@ import com.androidplot.pie.PieChart
 import com.androidplot.xy.XYPlot
 import com.futsch1.medtimer.R
 import com.futsch1.medtimer.database.MedicineRepository
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 
-class ChartsFragment : Fragment() {
+class ChartsFragment(
+    private val mainDispatcher: CoroutineDispatcher = Dispatchers.Main,
+    private val backgroundDispatcher: CoroutineDispatcher = Dispatchers.Default
+) : Fragment() {
     private lateinit var medicineRepository: MedicineRepository
 
     private lateinit var takenSkippedChartView: PieChart
@@ -55,7 +59,7 @@ class ChartsFragment : Fragment() {
         medicineRepository = MedicineRepository(requireActivity().application)
 
         setupTakenSkippedCharts()
-        lifecycleScope.launch(Dispatchers.Default) {
+        lifecycleScope.launch(backgroundDispatcher) {
             setupMedicinesPerDayChart()
             daysFlow.collect { days ->
                 populateStatistics(days)
@@ -96,12 +100,12 @@ class ChartsFragment : Fragment() {
 
         try {
             val series = statisticsProvider.getLastDaysReminders(days)
-            withContext(Dispatchers.Main) {
+            withContext(mainDispatcher) {
                 medicinesPerDayChart.updateData(series)
             }
 
             val data = statisticsProvider.getTakenSkippedData(days)
-            withContext(Dispatchers.Main) {
+            withContext(mainDispatcher) {
                 takenSkippedChart.updateData(
                     data.taken,
                     data.skipped,
@@ -110,7 +114,7 @@ class ChartsFragment : Fragment() {
             }
 
             val dataTotal = statisticsProvider.getTakenSkippedData(0)
-            withContext(Dispatchers.Main) {
+            withContext(mainDispatcher) {
                 takenSkippedTotalChart.updateData(
                     dataTotal.taken,
                     dataTotal.skipped,


ThomasKiljanczykDev:
I see there are failing tests here. I'll work on that

app/src/main/java/com/futsch1/medtimer/statistics/MedicinePerDayChart.kt
```diff
@@ -0,0 +1,244 @@
...
+            return buffer.append(daysSinceEpochToDateString(context, value.toLong()))
+        }
+
+        override fun format(
+            value: Long, buffer: StringBuffer, field: FieldPosition
+        ): StringBuffer {
+            throw UnsupportedOperationException("Not yet implemented.")
+        }
+
+        override fun parse(string: String, position: ParsePosition): Number? {
```
Futsch1, line 247: Make parse return a non-nullable type

app/src/main/java/com/futsch1/medtimer/statistics/MedicinePerDayChart.kt
```diff
@@ -0,0 +1,244 @@
...
+        formatter.fillPaint.setColor(getColor(title))
+        formatter.borderPaint.setColor(context.getMaterialColor(androidx.appcompat.R.attr.colorPrimary))
+        return formatter
+    }
+
+    private fun adjustTitleLength(series: SimpleXYSeries, columnPixels: Int, textPaint: Paint) {
+        var titleWidth = textPaint.measureText(series.title).toInt()
+        var title = series.title
+        while (titleWidth > columnPixels && title.length > 5) {
+            title = title.substring(0, title.length - 4) + "..."
```
Futsch1: use dropLast(4) instead of substring

Futsch1:
> I see there are failing tests here. I'll work on that

@ThomasKiljanczykDev This one test failing is a flaky one... I manage to iron out these flaky tests every now and then, but then a new issue pops up that randomly makes tests fail. It is usually no issue in the code, but some precondition on the simulated Android device (especially unpleasant are notifications that are already active in the simulated device and interfere with notification tests...). Still haven't found the magic bullet here to solve this.

app/src/main/java/com/futsch1/medtimer/helpers/ContextExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package com.futsch1.medtimer.helpers
+
+import android.content.Context
+import com.futsch1.medtimer.statistics.TakenSkippedChart
+import com.google.android.material.color.MaterialColors
+
+fun Context.getMaterialColor(attrId: Int): Int =
+    MaterialColors.getColor(this, attrId, TakenSkippedChart::class.simpleName)
```
Futsch1: Shouldn't we provide a generic string here instead of adding a dependency to a view?

app/src/main/java/com/futsch1/medtimer/helpers/ContextExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package com.futsch1.medtimer.helpers
+
+import android.content.Context
+import com.futsch1.medtimer.statistics.TakenSkippedChart
+import com.google.android.material.color.MaterialColors
+
+fun Context.getMaterialColor(attrId: Int): Int =
```
Futsch1: There is one more location where MaterialColors is used in statistics, ReminderTableFragment, line 55. It might make sense to also change this to the new extension function.

app/src/main/java/com/futsch1/medtimer/statistics/ChartsFragment.kt
```diff
@@ -0,0 +1,129 @@
...
+        val statisticsView = inflater.inflate(R.layout.fragment_charts, container, false)
+
+        medicinesPerDayChartView = statisticsView.findViewById(R.id.medicinesPerDayChart)
+        takenSkippedChartView = statisticsView.findViewById(R.id.takenSkippedChart)
+        takenSkippedTotalChartView =
+            statisticsView.findViewById(R.id.takenSkippedChartTotal)
+        medicineRepository = MedicineRepository(requireActivity().application)
+
+        setupTakenSkippedCharts()
+        lifecycleScope.launch(Dispatchers.Default) {
```
Futsch1: Do not use hardcoded dispatchers (https://developer.android.com/kotlin/coroutines/coroutines-best-practices?utm_source=android-studio-app&utm_medium=app#inject-dispatchers)

Futsch1: Same also later in this file using Dispatchers.Main

app/src/main/java/com/futsch1/medtimer/statistics/MedicinePerDayChart.kt
ThomasKiljanczykDev: Can' do. It's an override of a Java class method. All references in Java must be assumed nullable unless some not-null annotation is present.

This particular method throws `UnsupportedOperationException` so at least here this nullability is not an issue.

app/src/main/java/com/futsch1/medtimer/helpers/ContextExtensions.kt
ThomasKiljanczykDev: Right, it should be parameter here, and the name should be explicitly passed where it's used

app/src/main/java/com/futsch1/medtimer/helpers/ContextExtensions.kt
ThomasKiljanczykDev: I've applied it wherever I could.

`ReminderTableCellViewHolder` was a Java file, so I left a TODO there to not increase the scope of this PR

app/src/main/java/com/futsch1/medtimer/helpers/ContextExtensions.kt
ThomasKiljanczykDev: done

app/src/main/java/com/futsch1/medtimer/statistics/MedicinePerDayChart.kt
ThomasKiljanczykDev: done

app/src/main/java/com/futsch1/medtimer/statistics/ChartsFragment.kt
ThomasKiljanczykDev: I moved it to the constructor but it's of no value right now.
Hilt + a test DI module is needed to replace the dispatchers in any fragment tests.

Fragments are not unit-testable, so the constructor parameters are useless for now regardless if this project has proper DI set upped or not.

The example from docs shows a unit-testable class, so there it makes more sense.

app/src/main/java/com/futsch1/medtimer/statistics/MedicinePerDayChart.kt
Futsch1: Hm, the code checker in AndroidStudio reports an issue here...

app/src/main/java/com/futsch1/medtimer/helpers/ContextExtensions.kt
Futsch1: Ah, I overlooked that.

app/src/main/java/com/futsch1/medtimer/statistics/ChartsFragment.kt
Futsch1: I know the value is limited, but I'd like to keep the code clean of issues annotated by SonarQube or the AndroidStudio built-in linter. Reason is that the fewer you have, the easier it is to spot real issues. So even though it is not of actual value from a code perspective, it is from (my) developer's perspective.

app/src/main/java/com/futsch1/medtimer/statistics/MedicinePerDayChart.kt
Futsch1: Let's not be blocked by this one, I'll merge.

PR #1281, opened 2026-03-09
1260 kotlin flows

ThomasKiljanczykDev:
@Futsch1 please review this one.

This PR replaces all LiveData usages with flows.

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
@@ -27,7 +27,7 @@ open class MedicineRepository(val context: Context) {
         return medicineDao.getOnlyMedicine(medicineId)
     }
 
-    fun getFlowMedicine(medicineId: Int): Flow<FullMedicine?> {
+    fun getMedicineFlow(medicineId: Int): Flow<FullMedicine?> {
         return medicineDao.getMedicineFlow(medicineId)
     }
 
@@ -49,10 +49,6 @@ open class MedicineRepository(val context: Context) {
 
     fun getReminderFlow(reminderId: Int): Flow<Reminder?> {
         return medicineDao.getReminderFlow(reminderId)
-    }
-
-    fun getMedicineFlow(medicineId: Int): Flow<FullMedicine?> {
-        return medicineDao.getMedicineFlow(medicineId)
     }
 
     fun getReminderEventsFlow(timeStamp: Long, statusValues: List<ReminderStatus>): Flow<List<ReminderEvent>> {
--- a/app/src/main/java/com/futsch1/medtimer/overview/OverviewViewModel.kt
+++ b/app/src/main/java/com/futsch1/medtimer/overview/OverviewViewModel.kt
@@ -31,7 +31,7 @@ class OverviewViewModel(application: Application, medicineViewModel: MedicineVie
     private var _initialized = false
     val initialized get() = _initialized
 
-    private val _filterState = MutableStateFlow(FilterState(emptySet(), LocalDate.now(), 0L))
+    private val filterState = MutableStateFlow(FilterState(emptySet(), LocalDate.now(), 0L))
 
     private val reminderEvents =
         medicineViewModel.getLiveReminderEvents(Instant.now().toEpochMilli() / 1000 - (6 * 24 * 60 * 60), statusValuesWithoutDelete)
@@ -40,13 +40,13 @@ class OverviewViewModel(application: Application, medicineViewModel: MedicineVie
     private val sharedPreferences = PreferenceManager.getDefaultSharedPreferences(application)
 
     var day: LocalDate
-        get() = _filterState.value.day
+        get() = filterState.value.day
         set(value) {
-            _filterState.update { it.copy(day = value) }
+            filterState.update { it.copy(day = value) }
         }
 
     val overviewEvents: SharedFlow<List<OverviewEvent>> =
-        combine(reminderEvents, scheduledReminders, _filterState) { events, reminders, fs ->
+        combine(reminderEvents, scheduledReminders, filterState) { events, reminders, fs ->
             getFiltered(events, reminders, fs)
     }.onEach { _initialized = true }.shareIn(viewModelScope, SharingStarted.Eagerly, replay = 1)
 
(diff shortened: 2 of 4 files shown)


gradle.properties
```diff
@@ -6,22 +6,11 @@
...
-android.nonTransitiveRClass=true
+org.gradle.jvmargs=-Xmx4096m
+org.gradle.parallel=true
+org.gradle.caching=true
 org.gradle.configuration-cache=true
-android.uniquePackageNames=false
-android.dependency.useConstraints=false
-android.r8.strictFullModeForKeepRules=false
-# Disable the new DSL until the SonarQube plugin was updated
-android.newDsl=false
```
Futsch1, line 27: Disabling this now fully breaks the SonarQube plugin. But since it also does not work with AGP 9.1.0, we could now disable it. I can do this in a separate branch, but this will fail for sure on merge to main.

app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
```diff
@@ -21,23 +20,23 @@ open class MedicineRepository(val context: Context) {
...
+    val medicinesFlow: Flow<List<FullMedicine>>
+        get() = medicineDao.getMedicinesFlow()
 
     fun getOnlyMedicine(medicineId: Int): Medicine? {
         return medicineDao.getOnlyMedicine(medicineId)
     }
 
-    fun getLiveMedicine(medicineId: Int): LiveData<FullMedicine?> {
-        return medicineDao.getLiveMedicine(medicineId)
+    fun getFlowMedicine(medicineId: Int): Flow<FullMedicine?> {
```
Futsch1: Should be named `getMedicineFlow` - or was this intentional?

app/src/main/java/com/futsch1/medtimer/statistics/CalendarEventsViewModel.kt
```diff
@@ -115,14 +129,25 @@ class CalendarEventsViewModel(
                 continue
             }
 
-            val day = secondsSinceEpochToLocalDate(reminderEvent.remindedTimestamp, ZoneId.systemDefault())
-            if ((day >= startDay) && (medicine == null || medicine?.name == MedicineHelper.normalizeMedicineName(reminderEvent.medicineName))) {
-                eventListByDay.getOrPut(day) { mutableListOf() }.add(reminderEventToString(reminderEvent))
+            val day = secondsSinceEpochToLocalDate(
```
Futsch1, line 132: Do you have a different settings on formatting when it comes to line length? I prefer to have 160 chars, since it is easily viewable on today's screens and I prefer not to scroll down too much. No big deal, but this change here is actually none (if I'm not mistaken...)

app/src/main/java/com/futsch1/medtimer/MedicineViewModel.kt
```diff
@@ -14,72 +12,48 @@ import com.futsch1.medtimer.database.Tag
...
-    private lateinit var medicineToTags: List<MedicineToTag>
-    private val liveTags = medicineRepository.liveTags
+    private var medicineToTags: List<MedicineToTag> = emptyList()
+    private val liveTags: StateFlow<List<Tag>> = medicineRepository.tagsFlow.stateIn(
+        viewModelScope, SharingStarted.Eagerly, emptyList()
+    )
 
-    private val filteredMedicine = MediatorLiveData<List<FullMedicine>>()
-    val medicines: LiveData<List<FullMedicine>> = filteredMedicine
+    private val _scheduledReminders = MutableStateFlow<List<ScheduledReminder>>(emptyList())
```
Futsch1, line 37: I don't like the naming with the _ in the beginning. But I think this class could use some overhaul in terms of variable naming anyway, so we can also just keep it for the time being.

app/src/main/java/com/futsch1/medtimer/overview/OverviewViewModel.kt
```diff
@@ -19,32 +25,32 @@ enum class OverviewFilterToggles {
...
 
+data class FilterState(val activeFilters: Set<OverviewFilterToggles>, val day: LocalDate, val tick: Long)
+
 class OverviewViewModel(application: Application, medicineViewModel: MedicineViewModel) : AndroidViewModel(application) {
-    var initialized = false
-    val overviewEvents = MediatorLiveData<List<OverviewEvent>>()
+    private var _initialized = false
+    val initialized get() = _initialized
+
+    private val _filterState = MutableStateFlow(FilterState(emptySet(), LocalDate.now(), 0L))
```
Futsch1: I think here it's not necessary to put the _ in front of the name

app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
ThomasKiljanczykDev: I wonder if we need a flow for a single medicine.
Usually flows are used when fetching collections for UI lists since there it's better to observe the collection than having to decide when to fetch new data.

I will analyze if the previous LiveData was needed for anything sensible or was use just 'because'.

app/src/main/java/com/futsch1/medtimer/statistics/CalendarEventsViewModel.kt
ThomasKiljanczykDev: I'm using default settings provided by Android Studio so unless you changed the line width in the IDE, we should have the same settings

app/src/main/java/com/futsch1/medtimer/MedicineViewModel.kt
ThomasKiljanczykDev: `_` is the usual approach when having a mutable backing field for a immutable public field.
https://kotlinlang.org/docs/properties.html#backing-properties

I'd rather stick with this simple convention, then fallback to some naming gymnastics, or, god forbid, the old Hungarian Notation.

app/src/main/java/com/futsch1/medtimer/overview/OverviewViewModel.kt
ThomasKiljanczykDev: yep, not needed here

app/src/main/java/com/futsch1/medtimer/database/MedicineRepository.kt
ThomasKiljanczykDev: The current app architecture requires flows for these single entity calls.
This particular method was a naming mistake.
After changing it to a correct name it turned out it was a duplicate method.
I have removed it.

app/src/main/java/com/futsch1/medtimer/overview/OverviewViewModel.kt
ThomasKiljanczykDev: removed the `_`

gradle.properties
ThomasKiljanczykDev: I have revered the gradle.properties to the old settings but left the caching and memory changes

ThomasKiljanczykDev: It's best to check what are the default values, because I think some of these properties have the same defaults as set in the file and could be removed.

The `android.newDsl` will be needed moving forward.

app/src/main/java/com/futsch1/medtimer/statistics/CalendarEventsViewModel.kt
Futsch1: I actually did change the line width and will export the settings (although it was the only change)

app/src/main/java/com/futsch1/medtimer/MedicineViewModel.kt
Futsch1: In this case agree

PR #1294, opened 2026-03-13
Dependency injection of shared preferences

Futsch1:
@ThomasKiljanczykDev 
I started by only moving the weekend mode settings into an injected data source and putting the values in a separate model. While this worked well, testing is still a bit tricky and does not quite work in the corresponding unit test. So this PR is far from finished, but I wanted to get some early feedback before I waste too much time...

Edit: Unit test is fixed, so I do not have a HiltTest yet, but still done all the preconditions.

PR #1295, opened 2026-03-14
1285 repository full di migration part 1

ThomasKiljanczykDev:
@Futsch1 this one continues the work on using DI to inject repositories everywhere their used.
I have split it into arbitrary parts to limit the scope of individual PRs

There are a handful of instrumented tests changes.
I have encountered some hurdles there and spent bulk of my work on resolving them.
This includes changes to the notification tray open/closes handling and dialog handling.
I've also reduced the attempts from the default 10 to 3, because otherwise it took eternity for the tests to finish if a lot of them were failing.

### DIFF
--- a/app/src/androidTest/java/com/futsch1/medtimer/ReminderTest.kt
+++ b/app/src/androidTest/java/com/futsch1/medtimer/ReminderTest.kt
@@ -427,7 +427,7 @@ class ReminderTest : BaseTestHelper() {
             clickOn(R.id.takenButton)
 
             // Check if cyclic information is present
-            clickOn(R.id.overviewContentContainer)
+            clickListItemChild(R.id.reminders, 0, R.id.overviewContentContainer)
             if (reminder.shouldHaveInfo) {
                 assertContains(R.id.editEventName, String.format("Test (1/%d)", reminder.consecutiveDays))
             } else {
@@ -437,7 +437,7 @@ class ReminderTest : BaseTestHelper() {
             Espresso.pressBack()
 
             // Remove event
-            clickOn(R.id.stateButton)
+            clickListItemChild(R.id.reminders, 0, R.id.stateButton)
             clickOn(R.id.deleteButton)
             clickDialogPositiveButton()
 


Futsch1:
@ThomasKiljanczykDev 
Hm, I don't quite know what to say. On the one hand I am very grateful for your help, but I have created https://github.com/Futsch1/medTimer/pull/1294 yesterday with my first steps towards the DI domain. I know you told me to keep the scope small, but it eventually grew and included stuff like reminderContext, which is quite omnious in some parts of the app. Now I can think of either throwing it all away (since I certainly took some wrong turns, because I am still learning modern Kotlin) or try to merge... somehow.

We need to find a way on how to work together. I am not comfortable with letting you do all the work, because eventually I am responsible for the app and need to understand what it does and how it works.

So I suggest the following: You continue until a certain point and then stop and let me continue. I think it does not work to refactor in parallel.

Edit: Ok, I was able to rebase my branch on yours without issue... So maybe it does work after all. I will reconsider and see to it that we get both branches merged.

ThomasKiljanczykDev:
@Futsch1 your PR is smaller, so it has priority.

Removing the ReminderContext will be needed but it's omnipresent in the codebase, so your approach with the intermediate glue to make it work with Hilt at places is sound.

I could have made this PR as a draft and wait for the other to complete.

app/src/main/java/com/futsch1/medtimer/Autostart.kt
```diff
@@ -4,69 +4,41 @@ import android.content.BroadcastReceiver
...
-            }
-            Log.i(AUTOSTART, "Requesting reschedule")
-            requestScheduleNextNotification(context)
-        }
-    }
+@AndroidEntryPoint
+class Autostart : BroadcastReceiver() {
 
     companion object {
+        const val TAG = "Autostart"
```
Futsch1: I think we discussed this. I prefer to keep the log tags in a central file and manage them in one place... Even if this is not general convention... Same goes for the AutostartService.

ThomasKiljanczykDev: What are you using the tags for that you need the tag definition centralization?

According to the comment doc for logging, it should contain an identifier for a class/activity where the log is called.
There are already some logs in the app that break the log tag semantic convention which would have been avoided with the co-location.
image

The log tag semantics as described in the official docs of the log methods mean that each class should have its own unique log tag, therefore, is the tag's owner.
In this context the centralization just adds another layer of complexity to the development process, because you have to actively think of cleaning up the centralized tag constants after removing a class that used them.
Co-locating the tag with the class solves this problem.

Also imagine a situation where you start following the log tag semantics, so tag per class.
If you log from 50 classes, you have 50 centralized constants.
If you log from 100 classes, you have 100 centralized constants.
And so on.
This approach won't scale well.

This also reduces the velocity of coding since you have to go to the centralized file and add the tag there before proceeding with logging.
If the standard convention was followed, you would just slap a const val at the top of the file (or companion object for a class) and go on with the development.

Lastly, the standard conventions are there for a reason.
They reduce friction for new developers to enter a project.
If each project followed its own conventions, one would have to re-learn what they know each time they enter a new project.
There's also an impact on context switching if someone is working on multiple Android projects and need to constantly think which conventions are used in which project.

One would even argue that the centralized log tags class breaks the Single Responsibility principle of the SOLID principles, since you're adding a reason to change for centralized tags collection with each new tag.
The tag should reflect the class name, so each time you rename the class, you have to modify the centralized tags as well.
Same goes with adding/deleting the class that does the logging, another reason to change.


There are of course places where there's enough flexibility to use your own conventions, but logging is such an essential part of Android development that it would be better to follow what's recommended by the Android  documentation.

Futsch1: Let me ask differently: What are logs used for?

In my experience, they are used for debugging issues that occur at runtime. Issues at runtime usually do not happen in a single class, but are often involving multiple classes. But usually, they involve a single topic/functionality of an app. So if you want to set up a log filter to only see the involved classes with the non-centralized approach you would have to type in a bunch of class names. The centralized approach does not contain class names, but functionalities/modules (that's also why it scales - you only add new tags when you add a new functionality). In this case, it is only one constant to search for.

I often had to debug issues related to reminder scheduling. That's why all classes in the reminders package share a single log tag, because they work together and implement a flow of actions. I can easily filter and check these. When I debug, I know which function does not work properly, I do NOT know which class doesn't.

So please: Keep this principle as it is.

app/src/main/java/com/futsch1/medtimer/overview/EditEventSheetDialogFragment.kt
Futsch1, line 1: This fragment seems fairly complicated... Would it be worth thinking about introducing Compose right here? You have a working ViewModel already. I know we had this later in the migration step, but it might be worth considering if the complexity would be reduced with such a move.

app/src/main/java/com/futsch1/medtimer/Autostart.kt
ThomasKiljanczykDev: For now I have reverted back to LogTags.

That being said, the debugging flow you described is better handled by plain old debugger with breakpoints or tracing with log correlation.

There's an OpenTelemetry integration for Android that would handle that well, but Android Studio does not have in-built OpenTelemetry collector like in JetBrains Rider.
It would complicate the stack for now.
Maybe JetBrains will expand this extension support to all IDEs in the future.

app/src/main/java/com/futsch1/medtimer/overview/EditEventSheetDialogFragment.kt
ThomasKiljanczykDev: I hoped to first the DI work completed, then tidy remaining architectural missteps, then restart the compose effort.

At bare minimum lets have the DI work finished so the compose PRs focus more on the UI stuff than overall refactoring.

app/src/main/java/com/futsch1/medtimer/medicine/EditMedicineFragment.kt
```diff
@@ -153,25 +157,39 @@ class EditMedicineFragment : Fragment(), IconDialog.Callback {
...
+    private fun onEntityLoaded(entity: FullMedicine, fragmentView: View) {
         val medicine = entity.medicine
 
         color = medicine.color
         iconId = medicine.iconId
         notes = medicine.notes
 
         (requireActivity() as AppCompatActivity).supportActionBar?.title = medicine.name
-        fragmentView.findViewById<EditText>(R.id.editMedicineName).setText(medicine.name)
+        fragmentView.findViewById<EditText>(R.id.editMedicineName).apply {
```
Futsch1, line 179: Why did you add this? The idea of this fragment (before) was to only update the entity (aka medicine) on leaving the fragment. So in fillEntityData() this should be done and this after text change handling is not required. (Not saying that it should stay like this after the compose migration, but that's at least how the original code intended to do it).

app/src/main/java/com/futsch1/medtimer/overview/EditEventSheetDialogFragment.kt
Futsch1: Fine for me, was just wondering.

Futsch1:
@ThomasKiljanczykDev 
I restarted the one failing test - let's see if it is a flaky one. As soon as all tests are passed and my PR #1296 is merged, yours is ready to follow.

Futsch1:
@ThomasKiljanczykDev 
The SDK 28 tablet tests have a systematic issue. You can also download the reports of the failed job, they contain screenshots, logs and UI dumps of the failed tests.

app/src/main/java/com/futsch1/medtimer/reminders/AlarmProcessor.kt
```diff
@@ -64,7 +45,26 @@
                 )
             )
             ReminderNotificationProcessor(reminderContext).processReminders(scheduledReminderNotificationData)
+            return
+        }
+
+        val pendingIntent = scheduledReminderNotificationData.getPendingIntent(reminderContext)
+
+        if (canScheduleExactAlarms()) {
+            alarmManager.setExactAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, scheduledReminderNotificationData.remindInstant.toEpochMilli(), pendingIntent)
```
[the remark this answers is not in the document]
ThomasKiljanczykDev: Should be addressed with my latest commit

app/src/main/java/com/futsch1/medtimer/reminders/AlarmProcessor.kt
```diff
@@ -64,7 +45,26 @@
...
+        } else {
+            alarmManager.setAndAllowWhileIdle(AlarmManager.RTC_WAKEUP, scheduledReminderNotificationData.remindInstant.toEpochMilli(), pendingIntent)
```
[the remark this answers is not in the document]
ThomasKiljanczykDev: Should be addressed with my latest commit

ThomasKiljanczykDev:
@Futsch1 this PR should be ready for final review/approval.

Please re-run the failed checks if you can.

I have implemented some test reliability changes so the tablet (API 28) tests should not fail.
They passed in the previous pipeline run.

app/src/androidTest/java/com/futsch1/medtimer/NotificationTest.kt
```diff
@@ -194,7 +197,7 @@ class NotificationTest : BaseTestHelper() {
     }
 
     @Test
-    //@AllowFlaky(attempts = 1)
+    @Ignore("Repeat timing unreliable with exact reminders disabled")
```
Futsch1: I would prefer to either explore how to enable exact reminders in the tests or to extend the timeout from 120 seconds to 180 seconds. Ignoring the test and therefore reducing test coverage does not seem a good idea. I know that these tests are flaky at times, but they helped finding bugs early in the past...

I would be ok with marking this one as //TODO now and revisiting it later.


```suggestion
    // TODO: Enable exact reminders in this test if possible or extend the timeout.
    @Ignore("Repeat timing unreliable with exact reminders disabled")
```

ThomasKiljanczykDev: There was a bug in the app, where the repeated notification was not happening when using the combined reminder notifications and marking one medicine as taken.
I've fixed it.

I've also enabled the 'exact reminders' to make the test less flaky.

Futsch1: Great job, thank you. I will review and merge

PR #1324, opened 2026-03-24
Di medicine repository

ThomasKiljanczykDev:
@Futsch1 this one eliminates last usages of direct MedicineRepository constructor

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/OptionsMenu.kt
+++ b/app/src/main/java/com/futsch1/medtimer/OptionsMenu.kt
@@ -277,7 +277,7 @@ class OptionsMenu @AssistedInject constructor(
             }
         }
         val medicines: List<FullMedicine> = medicineViewModel.filterMedicines(medicineViewModel.medicineRepository.medicines)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/futsch1/medtimer/overview/ShowMedicineViewModel.kt
```diff
@@ -0,0 +1,74 @@
...
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+sealed interface ShowMedicineUiState {
+    data object Loading : ShowMedicineUiState
+    data object NotFound : ShowMedicineUiState
+    data class Loaded(
+        val fullMedicine: FullMedicine,
+        val reminder: Reminder,
+        val reminderSummaryText: String,
```
Futsch1, line 28: I was wondering, is it correct to calculate string representation in the view model? Since this might depend on how the view is built, stuff like string length, line breaks etc might be dependent on the view. I also do this in another View model (calendar), where I would now decide differently and have the view decide how to render reminder strings. 

@ThomasKiljanczykDev what is your opinion?

PR #1330, opened 2026-03-28
DI remaining system services

ThomasKiljanczykDev:
@Futsch1 I've found a usages of system services which could have been migrated to DI.

These changes had some cascading consequences, so there are even more classes that use DI/assisted DI now.

### DIFF
--- a/app/src/androidTest/java/com/futsch1/medtimer/NotificationTest.kt
+++ b/app/src/androidTest/java/com/futsch1/medtimer/NotificationTest.kt
@@ -394,9 +394,8 @@ class NotificationTest : BaseTestHelper() {
             dismissNotification(notification)
         }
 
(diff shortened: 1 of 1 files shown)


app/src/androidTest/java/com/futsch1/medtimer/NotificationTest.kt
```diff
@@ -394,8 +394,9 @@ class NotificationTest : BaseTestHelper() {
             dismissNotification(notification)
         }
 
-        device.wait(Until.findObject(By.displayId(android.R.id.input)), 2_000)
-        writeTo(android.R.id.input, "5")
+        val input = device.wait(Until.findObject(By.res("android", "input")), 2_000)
+        assertNotNull(input)
+        input.text = "5"
```
ThomasKiljanczykDev: Should be reverter

Futsch1 (approved):

Issue #1339, opened 2026-04-01
Support for intervals over one month


Jasper-Mare:
I have a medicine which my doctor administers every 3 months, I want to add a reminder to book the appointment.

Unfortunately, the continuous interval reminder only allows a maximum interval of 31 days (3 months is 91 days)

I would either like for the cap to be raised, or for a "months" interval option to be added.

jennifer-mare:
PR:
[https://github.com/Futsch1/medTimer/pull/1340](https://github.com/Futsch1/medTimer/pull/1340)

PR #1340, opened 2026-04-01
Issue-1339

jennifer-mare:
issue:
[https://github.com/Futsch1/medTimer/issues/1339](https://github.com/Futsch1/medTimer/issues/1339)

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/helpers/Interval.kt
+++ b/app/src/main/java/com/futsch1/medtimer/helpers/Interval.kt
@@ -4,13 +4,14 @@ import android.content.Context
 import com.futsch1.medtimer.R
 
 enum class IntervalUnit {
-    MINUTES, HOURS, DAYS
+    MINUTES, HOURS, DAYS, MONTHS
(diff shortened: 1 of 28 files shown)


jennifer-mare:
I've set the values to ??? for the translations because I'm not confident about translating all these languages correctly

jennifer-mare:
I'm not sure how to edit the ui, but all the backend code should be done (keep in mind I'm not familiar with this codebase so I've probably missed something)

jennifer-mare:
I'm not sure how to compile and run tests, should I have android studio?

Futsch1:
@Jasper-Mare 
Thanks for your contribution! Yes, compilation and tests can be performed in Android Studio.

But before you get started: using months as additional unit would lead to confusion I fear. Imagine setting to 3 months and internally calculating with days: this would result in a drift of the day of the month, which might not be what users expect. I would rather increase the cap to 100 days instead.

Futsch1:
Closing this MR as it is superseded by #1384

PR #1343, opened 2026-04-03
Split repositories and DAOs into entity-specific classes, update repository/DAO method naming to not be entity-dependent

ThomasKiljanczykDev:
@Futsch1 I've already finished this one while waiting for di-final merge.

This PR splits the single repository and DAO into per-entity classes to make them smaller and easier to manage (as well as adhere to Interface Segregation and Single Responsibility principles yada yada yada).

You can decide if you'll want to merge this one first or your domain model abstractions.
I didn't know you started implementing the domain abstraction already when was doing the repository split.


ThomasKiljanczykDev:
@Futsch1 there are some false-positives in the linter.
I can easily build and run the unit tests locally.
The lint also passes locally.

On the side note - I would recommend to disable `warningsAsErrors` in the lint config. This option may lead to dangerous situations where you are forced to update to latest dependency libraries versions. This may lead to 2 situations: updating to a unstable version with known bugs, update to a library version that has been compromised (i.e. hacked and modified with malware). If you need specific rules to be treated as errors, you can add them to the error or fatal collection in the config.

Futsch1:
I would prefer to do it the other way around: exclude the version warnings of both the SDK and the libraries (this is already active for branches) and rely on Dependabot.

PR #1347, opened 2026-04-06
Refactoring to use model classes instead of database entitites

Futsch1:


Futsch1:
@ThomasKiljanczykDev 
I have drafted the PR to use only model classes in the application code. It is also a big one, but the scope of changes is limited to this (and a few minor things left and right). Please have a look.

Android Tests still fail, so there's a few bugs left which I will fix.

app/src/main/java/com/futsch1/medtimer/database/backup/BackupManager.kt
```diff
@@ -52,10 +52,10 @@ class BackupManager @AssistedInject constructor(
     @Assisted private val fragmentManager: FragmentManager? = null,
     private val preferencesDataSource: PreferencesDataSource,
     private val persistentDataDataSource: PersistentDataDataSource,
-    private val medicineRepository: MedicineRepository,
-    private val reminderRepository: ReminderRepository,
-    private val reminderEventRepository: ReminderEventRepository,
-    private val tagRepository: TagRepository,
+    private val medicineDao: MedicineDao,
```
ThomasKiljanczykDev, line 55: Why have you replaced repositories with dao?
The repositories were meant to abstract ALL the code from the db implementation.

With this mixed approach you're breaking this layer of abstraction

app/src/main/java/com/futsch1/medtimer/database/backup/BackupManager.kt
```diff
@@ -199,7 +199,7 @@ class BackupManager @AssistedInject constructor(
     }
 
     private fun createAndSave(fileContent: String?) {
-        val file = File(context.cacheDir, backupFilename)
+        val file = File(context.cacheDir, PathHelper.backupFilename)
```
ThomasKiljanczykDev: Helper (and Util) is a very vague name which often leads to dumping a lot of unrelated code into a single class.
It would probably be better to give it some more sensible name

app/src/main/java/com/futsch1/medtimer/model/Medicine.kt
```diff
@@ -0,0 +1,61 @@
...
+
+import com.futsch1.medtimer.ReminderNotificationChannelManager
+import java.time.LocalDate
+
+data class Medicine(
+    val name: String,
+    val id: Int,
+    val color: Int,
+    val useColor: Boolean,
+    val notificationImportance: ReminderNotificationChannelManager.Importance,
```
ThomasKiljanczykDev: You should probably extract the importance into `NotificationImportance` to make it independent from the channel manager. Channels themself are more part of the application layer

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
```diff
@@ -0,0 +1,73 @@
...
+import java.time.Instant
+
+class JSONMedicineBackup(
+    private val medicineDao: MedicineDao,
+    private val reminderDao: ReminderDao,
+    private val tagDao: TagDao
+) : JSONBackup<FullMedicineEntity>(FullMedicineEntity::class.java) {
+    override fun createBackup(databaseVersion: Int, list: List<FullMedicineEntity>): JsonElement {
+        // Correct the medicines where the reminder's instructions are null in this loop
+        for (fullMedicine in list) {
```
ThomasKiljanczykDev: You could use flatMap here instead of this for

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
```diff
@@ -0,0 +1,73 @@
...
+            }
+            val medicineId = medicineDao.insertMedicine(fullMedicine.medicine)
+            processReminders(fullMedicine, medicineId.toInt())
+            processTags(fullMedicine, medicineId.toInt())
+        }
+    }
+
+    private suspend fun processTags(fullMedicine: FullMedicineEntity, medicineId: Int) {
+        for (tag in fullMedicine.tags) {
+            val tagId = tagDao.create(tag).toInt()
```
ThomasKiljanczykDev, line 57: You should probably sort out this ID casting nonsense.

It should be either `int` everywhere or `long` everywhere.
`Long` is the safer option since the performance impact should not be that noticeable and you get practically infinite possible values

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
```diff
@@ -0,0 +1,73 @@
...
+            tagDao.createMedicineToTag(MedicineToTagEntity(medicineId, tagId))
+        }
+    }
+
+    private suspend fun processReminders(fullMedicine: FullMedicineEntity, medicineId: Int) {
```
ThomasKiljanczykDev, line 62: You should accept the domain model and copy the reminders (with id reset)

app/src/main/java/com/futsch1/medtimer/database/MedicineEntity.kt
```diff
@@ -7,11 +7,9 @@ import androidx.room.Ignore
 import androidx.room.PrimaryKey
 import com.futsch1.medtimer.ReminderNotificationChannelManager
 import com.google.gson.annotations.Expose
-import java.time.LocalDate
-import java.util.Objects
 
-@Entity
-class Medicine() {
+@Entity(tableName = "Medicine")
+class MedicineEntity() {
```
ThomasKiljanczykDev: You could move all the properties into the constructor so you create new entities more easily without for default values

ThomasKiljanczykDev (changes requested):
I haven't reviewed everything, but some of these comment are disruptive enough that the code might change a lot.

I will continue the review once you address my current comments.

app/src/main/java/com/futsch1/medtimer/database/backup/BackupManager.kt
Futsch1: I have moved this class into the database package for a reason: I consider the backup tightly coupled to the database implementation, so it is a database backup, not a model backup (also for the sake of compatibility). Since repositories interface is now based on the model classes, the backup requires a direct database access.

So I do not consider this breaking the abstraction.

app/src/main/java/com/futsch1/medtimer/database/backup/BackupManager.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/model/Medicine.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
Futsch1: See my comment above: I'd like to keep the backup logic close to the database structure.

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
Futsch1: I do not see how: Room always returns long from @Insert operations. I chose to use int as ID type some time ago and do not want to do a migration just to change this to long.

The repositories now always return int IDs, so towards the application it's int only. Internally, we have to do a cast, but I do not see how this hurts.

app/src/main/java/com/futsch1/medtimer/database/MedicineEntity.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
Futsch1: Done

Futsch1:
@ThomasKiljanczykDev 
I have addressed all your review comments

Futsch1:
@ThomasKiljanczykDev 
Can you let me know when you'll be able to review? No hurries, but still would be good to know...

ThomasKiljanczykDev:
@Futsch1 I've been tied up by other stuff.
I plan to complete the review next week, probably in the second half of the week.

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
ThomasKiljanczykDev: You're casting a type of higher precision (`long`) to a type of lower precision (`int`).
There's a (probably not that likely) case where some user will insert so many entities that the number will overflow and there'll be some entities that can't have IDs mapped to `int` and then back to `long`.

The cast is superficial to begin with and does not seem to bring any real value.
The `long` should be preserved everywhere without the casting.

Maybe it would be best to file a separate issue just for that.

app/src/main/java/com/futsch1/medtimer/database/backup/BackupManager.kt
ThomasKiljanczykDev: I would disagree.

Backup is most probably an application layer logic that orchestrates multiple data sources and either imports a backup to them or exports a backup from them.

The whole point of repositories is to abstract away from specific db implementation so you can replace the db engine without modifying the rest of the code.
Production replacement probably won't happen in a long time, but it's of great value for true class isolation when writing unit tests.

The backup logic itself should be oblivious to what is the data source, it should receive a contract (repository interface) and do the backup logic.

The fact that this class depends on other data sources and android construct (fragment manager) indicates it's not part of the 'core' database 'module'.

Probably the most valid approach would be to create a `BackupRepository` which exposes all the backup-specific database operations (CRUD) and abstract all the backup logic from the specific Room database backend.


That being said, reworking the whole backup logic for proper separation would be too much work in this PR.
Better to go back to this discussion later when more important refactoring tasks are all done.

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
ThomasKiljanczykDev: I admit I don't remember all the technicalities of Room.
I see that Room always returns Long even the model has an Int ID type.

Most cases of `.toInt()` are just leftovers from the direct DAO usages that have been replaced by the repositories.

So I guess there's no issue here.

app/src/main/java/com/futsch1/medtimer/exporters/CSVEventExport.kt
```diff
@@ -42,19 +42,19 @@ class CSVEventExport @AssistedInject constructor(
...
                             "%s;%s;%s;%s;%s;%s;%s;%s;%s\n",
                             timeFormatter.secondsSinceEpochToDateTimeString(
-                                reminderEvent.remindedTimestamp
+                                reminderEvent.remindedTimestamp.epochSecond
                             ),
                             reminderEvent.medicineName,
                             reminderEvent.amount,
                             if (reminderEvent.status == ReminderEvent.ReminderStatus.TAKEN) timeFormatter.secondsSinceEpochToDateTimeString(
-                                reminderEvent.processedTimestamp
+                                reminderEvent.processedTimestamp.epochSecond
```
ThomasKiljanczykDev, line 50: Seems like a candidate of a new timeFormatter function that accepts an `Instant` directly.

app/src/main/java/com/futsch1/medtimer/database/backup/JSONMedicineBackup.kt
ThomasKiljanczykDev: This discussion is closed

app/src/main/java/com/futsch1/medtimer/helpers/ReminderHelper.kt
```diff
@@ -1,33 +1,36 @@
...
 
 suspend fun setRemindersActive(reminders: List<Reminder>, reminderRepository: ReminderRepository, active: Boolean) {
-    for (reminder in reminders) {
-        if (!reminder.active && active && reminder.reminderType == Reminder.ReminderType.CONTINUOUS_INTERVAL) {
+    val updated = reminders.map { reminder ->
+        var r = reminder
+        if (!reminder.active && active && reminder.reminderType == ReminderType.CONTINUOUS_INTERVAL) {
             // If reminder is activated again and an interval reminder, reset the interval start date to the current day in seconds since epoch
-            reminder.intervalStart = TimeHelper.changeTimeStampDate(reminder.intervalStart, LocalDate.now())
+            r = r.copy(intervalStart = Instant.ofEpochSecond(TimeHelper.changeTimeStampDate(reminder.intervalStart.epochSecond, LocalDate.now())))
```
ThomasKiljanczykDev, line 31: `changeTimeStampDate` is candidate to accept `Instant` directly

app/src/main/java/com/futsch1/medtimer/helpers/ReminderStringFormatter.kt
```diff
@@ -27,20 +27,18 @@ class ReminderStringFormatter @Inject constructor(
 ) {
     fun formatReminderEvent(reminderEvent: ReminderEvent): Spanned {
         var takenTime = timeFormatter.secondsSinceEpochToConfigurableTimeString(
```
ThomasKiljanczykDev, line 29: This is a candidate to accept `Instant` directly

app/src/main/java/com/futsch1/medtimer/helpers/ReminderStringFormatter.kt
```diff
@@ -135,19 +125,28 @@ class ReminderStringFormatter @Inject constructor(
...
+
+    private fun getReminderTypeSpan(reminderEvent: ReminderEvent): Spanned =
+        getReminderTypeSpan(reminderEvent.reminderType)
+
+    private fun getLastIntervalTime(reminderEvent: ReminderEvent): String {
+        val lastIntervalTime = reminderEvent.lastIntervalReminderTimeInMinutes
+        val processedTimestamp = reminderEvent.processedTimestamp.epochSecond
+        val durationMillis = processedTimestamp * 1000L - lastIntervalTime * 60_000L
+        return if (lastIntervalTime > 0 && reminderEvent.status == ReminderEvent.ReminderStatus.TAKEN && durationMillis >= 0) {
+            " (" + context.getString(R.string.interval_time, formatDuration(durationMillis).toString()) + ")"
```
ThomasKiljanczykDev, line 137: Maybe https://kotlinlang.org/api/kotlinx-datetime/kotlinx-datetime/kotlinx.datetime/format.html this would work for date formatting?

app/src/main/java/com/futsch1/medtimer/medicine/editMedicine/NotificationImportance.kt
```diff
@@ -1,33 +1,2 @@
 package com.futsch1.medtimer.medicine.editMedicine
```
ThomasKiljanczykDev, line 1: From what I see, this file is empty and can be deleted

app/src/main/java/com/futsch1/medtimer/medicine/EditMedicineFragment.kt
```diff
@@ -107,7 +103,7 @@ class EditMedicineFragment : Fragment(), IconDialog.Callback {
     @Inject
     lateinit var reminderRepository: ReminderRepository
 
-    private var entity: FullMedicine? = null
+    private lateinit var medicine: Medicine
```
ThomasKiljanczykDev, line 106: It would probably be best to leave it as nullable, since this variable is filled in a asynchronous manner.
A nullable type will ensure that all possible error places are revealed by the linter and during the compilation.

In the current setup you can get an error, e.g. when stop is called before the medicine is fetched.

The number of null checks can be reduced by passing a pre-checked variable into the functions as parameter

app/src/main/java/com/futsch1/medtimer/medicine/LinkedReminderAlgorithms.kt
```diff
@@ -1,29 +1,31 @@
...
-        return reminders.sortedBy { r -> getTotalTimeInMinutes(r, reminders) }
+        return reminders.sortedBy { r -> getTotalTimeInSeconds(r, reminders) }
     }
 
-    private fun getTotalTimeInMinutes(reminder: Reminder, reminders: List<Reminder>): Int {
-        var total = if (reminder.reminderType != Reminder.ReminderType.WINDOWED_INTERVAL) reminder.timeInMinutes else reminder.intervalStartTimeOfDay
+    private fun getTotalTimeInSeconds(reminder: Reminder, reminders: List<Reminder>): Long {
+        var total =
+            (if (reminder.reminderType != ReminderType.WINDOWED_INTERVAL) reminder.time.seconds else reminder.intervalStartTimeOfDay.toSecondOfDay().toLong())
         for (r in reminders) {
```
ThomasKiljanczykDev, line 17: Could probably be replaced with
```kotlin
total += reminders                                                                                                                                                                                                
      .filter { it.id == reminder.linkedReminderId }                                                                                                                                                                
      .sumOf { r ->                                                                                                                                                                                                 
          when (r.reminderType) {
              ReminderType.LINKED -> getTotalTimeInSeconds(r, reminders)                                                                                                                                            
              else -> r.time.seconds
          }                                                                                                                                                                                                         
      } 
```

app/src/main/java/com/futsch1/medtimer/model/ReminderTime.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.time.LocalTime
+
+class ReminderTime : Comparable<ReminderTime> {
+    val minutes: Int
+    val seconds: Long
+        get() = minutes * 60L
+
+    val isDuration: Boolean
+
+    constructor(minutes: Int, isDuration: Boolean = false) {
```
ThomasKiljanczykDev, line 12: Could this be used as the primary constructor (i.e. specified by the class name) and the other constructor could call this one?

app/src/main/java/com/futsch1/medtimer/overview/actions/ScheduledReminderActions.kt
```diff
@@ -60,7 +60,7 @@ class ScheduledReminderActions @AssistedInject constructor(
 
     private fun scheduleReminder(scheduledReminder: ScheduledReminder) {
         timePickerDialogFactory
-            .create(scheduledReminder.reminder.timeInMinutes / 60, scheduledReminder.reminder.timeInMinutes % 60) { minutes ->
+            .create(scheduledReminder.reminder.time.getLocalTime().hour, scheduledReminder.reminder.time.getLocalTime().minute) { minutes ->
```
ThomasKiljanczykDev, line 63: Maybe there could be a `create` function overload that accepts local time and passes the relevant hour and minute to the main implementation there?

app/src/main/java/com/futsch1/medtimer/reminders/ReminderNotificationProcessor.kt
```diff
@@ -112,69 +111,79 @@ class ReminderNotificationProcessor @Inject constructor(
...
         private suspend fun getLastReminderEventTimeInMinutes(
             reminderEventRepository: ReminderEventRepository,
-            reminderEvent: ReminderEvent,
+            reminderId: Int,
+            remindedTimestamp: Instant,
             isWindowedInterval: Boolean
         ): Int {
-            val lastReminderEvent = reminderEventRepository.getLast(reminderEvent.reminderId)
+            val lastReminderEvent = reminderEventRepository.getLast(reminderId)
             return if (lastReminderEvent != null && lastReminderEvent.status == ReminderEvent.ReminderStatus.TAKEN) {
```
ThomasKiljanczykDev, line 174: This if is quite nested.
Maybe it would be possible to invert the conditions and return `0` via explicit `return` to make this part more readable

app/src/main/java/com/futsch1/medtimer/reminders/RepeatProcessor.kt
```diff
@@ -25,7 +25,6 @@ class RepeatProcessor @Inject constructor(
 
     private suspend fun decreaseRemainingRepeats(reminderEventId: Int) {
         val reminderEvent = reminderEventRepository.get(reminderEventId) ?: return
-        reminderEvent.remainingRepeats -= 1
-        reminderEventRepository.update(reminderEvent)
+        reminderEventRepository.update(reminderEvent.copy(remainingRepeats = reminderEvent.remainingRepeats - 1))
```
ThomasKiljanczykDev, line 28: There is a risk of race condition here.
Maybe you could implement a transaction secured implementation in the DAO (similar to stock decrease)

app/src/main/java/com/futsch1/medtimer/reminders/StockHandlingProcessor.kt
```diff
@@ -17,18 +17,18 @@ class StockHandlingProcessor @Inject constructor(
...
+        Log.d(LogTags.STOCK_HANDLING, "Decrease stock for medicine ${medicine.name} by $amount resulting in ${medicine.amount}.")
+        checkForThreshold(medicine, amount, processedInstant)
     }
 
-    private fun checkForThreshold(fullMedicine: FullMedicine, decreaseAmount: Double, processedInstant: Instant) {
-        for (reminder in fullMedicine.reminders) {
-            if (reminder.reminderType == Reminder.ReminderType.OUT_OF_STOCK && fullMedicine.medicine.amount <= reminder.outOfStockThreshold) {
+    private fun checkForThreshold(medicine: Medicine, decreaseAmount: Double, processedInstant: Instant) {
+        for (reminder in medicine.reminders) {
+            if (reminder.outOfStockReminderType != Reminder.OutOfStockReminderType.OFF && medicine.amount <= reminder.outOfStockThreshold) {
```
ThomasKiljanczykDev, line 27: Could invert the `if` (with `continue`) to reduce the code width

ThomasKiljanczykDev (commented):
@Futsch1 I've completed the review.

I've limited my comments to mostly what has been changed to not blow out the scope even more.

There's not that much comments.
I'll review the previous comments how they've been addressed and will resolve them.

EDIT: GitHub fails when I try to resolve the comments. I'll leave the be

app/src/main/java/com/futsch1/medtimer/reminders/StockHandlingProcessor.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/reminders/RepeatProcessor.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/reminders/ReminderNotificationProcessor.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/overview/actions/ScheduledReminderActions.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/model/ReminderTime.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/exporters/CSVEventExport.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/ReminderHelper.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/helpers/ReminderStringFormatter.kt
Futsch1: Let's do this in a later step

app/src/main/java/com/futsch1/medtimer/medicine/editMedicine/NotificationImportance.kt
Futsch1: Done

app/src/main/java/com/futsch1/medtimer/medicine/EditMedicineFragment.kt
Futsch1: Done

Futsch1:
@ThomasKiljanczykDev I have commented or addressed al I'll review comments

ThomasKiljanczykDev (approved):

app/src/main/java/com/futsch1/medtimer/medicine/LinkedReminderAlgorithms.kt
Futsch1: I think I forgot this one, but I will not let this block the merge

PR #1410, opened 2026-05-13
Potential fix for timer repeat notification

gfotiades:
Cases where repeat count is checked may not handle 0/-1 difference correctly.  Have not verified.

### DIFF
--- a/app/src/main/java/com/futsch1/medtimer/database/dao/ReminderEventDao.kt
+++ b/app/src/main/java/com/futsch1/medtimer/database/dao/ReminderEventDao.kt
@@ -58,7 +58,9 @@ interface ReminderEventDao {
     @Transaction
     suspend fun decreaseRepeats(reminderEventId: Int) {
         val reminderEvent = get(reminderEventId) ?: return
-        reminderEvent.remainingRepeats = maxOf(0, reminderEvent.remainingRepeats - 1)
-        update(reminderEvent)
+        if (reminderEvent.remainingRepeats > 0) {
+            reminderEvent.remainingRepeats -= 1
+            update(reminderEvent)
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/futsch1/medtimer/database/dao/ReminderEventDao.kt
```diff
@@ -58,7 +58,9 @@ interface ReminderEventDao {
     @Transaction
     suspend fun decreaseRepeats(reminderEventId: Int) {
         val reminderEvent = get(reminderEventId) ?: return
-        reminderEvent.remainingRepeats = maxOf(0, reminderEvent.remainingRepeats - 1)
-        update(reminderEvent)
+        if (reminderEvent.remainingRepeats > 0) {
```
Futsch1, line 61: This is actually the same than maxOf(0, reminderEvent.remainintRepeats - 1). So this would not fix it.

Futsch1 (changes requested):
This change does not fix the issue unfortunately.

app/src/main/java/com/futsch1/medtimer/database/dao/ReminderEventDao.kt
gfotiades: Assuming `remainingRepeats = -1` (unlimited):
```
maxOf(0, -2) -> 0
remainingRepeats = 0
update(0) // repeats disabled
```

vs
```
remainingRepeats = -1
<no decrement, remainingRepeats stays at -1>
```

Was it wrong assumption that `-1` is unlimited?

Futsch1: I just saw this as well, I missed the unlimited reminders case. You are right, this is definitely a bug and your changes fix it! I will merge your fix, thanks a lot!

Futsch1 (approved):

Futsch1:
> This change does not fix the issue unfortunately.

It does, I was mistaken!

app/src/main/java/com/futsch1/medtimer/database/dao/ReminderEventDao.kt
gfotiades: Glad to hear it and glad I could help!  I like the app a lot.  I live by unlimited reminders.  Hearing the ding every minute across the room forever is annoying enough to never miss my meds, but it doesn't blaringly interrupt everything this very second like an alarm does.  Works great if I'm in a meeting or something.

PR #1419, opened 2026-05-18
Add support for supply tracking

### LINKED ISSUE #1205: Tracking and Option for more than one Packages
Some Medis or Nems have multiple Packages. For example one Package, Botles, ... are more than ones available.
One Package, Botle, ... have 78 Capseln, Tablets, ... and its 2, 3, ... in the Stock. Summary have 2 Packages, Botles, ... with Summary 156 Capseln, Tablets, ...

We have only an Option for a Package Size not an extended Option for a Number of Packages, Bootles, ...

A Option for Number of Packages, Bootles, ... in Stock.

BlankEclair:
[1.webm](https://github.com/user-attachments/assets/e60dc643-721a-4758-b0b3-1c097ead8951)

Fixes #1205


BlankEclair:
(I don't know how to fix the lint failure, because I can't quite translate all the strings myself)

Futsch1:
@BlankEclair 
Thanks a lot for the contribution and for working on a new feature!

I am however not yet fully convinced how this addition is really helpful. From my point of view, this is now a bit confusing: The total amount does not say how many units of the medicine I have at my disposal, but only how many are left in my current box/"supply". These two run-out dates are also somewhat confusing, because it is not immediately obvious how these numbers relate.

I receive a lot of feedback about the app being too complex and not user friendly and I think this feature needs some overhaul in terms of clarity and simplicity if it were to be added.

What do you think, can you understand?

BlankEclair:
(Sorry for the late reply, life is busy)

Yeah, that makes sense. I'll close this for now since I don't have the time to think much about it (and I did make a Python script to calculate it for me in the end)

PR #1435, opened 2026-05-22
Bump SonarQube plugin version and use new ApplicationExtension from A…

Futsch1:
…GP 9.0


core/database/src/main/java/com/futsch1/medtimer/database/MedicineRepositoryImpl.kt
```diff
@@ -11,7 +11,7 @@ import kotlinx.coroutines.flow.map
 class MedicineRepositoryImpl(
     private val medicineDao: MedicineDao
 ) : MedicineRepository {
-    override suspend fun get(medicineId: Int): Medicine? {
+    override suspend operator fun get(medicineId: Int): Medicine? {
```
ThomasKiljanczykDev, line 14: @Futsch1 I don't think using the `operator` for get in repositories is a good idea.

This particular `get` is not a retrieval from a materialized collection as the `operator get` is normally used for.

Here each `operator get` executes a query and returns a freshly-loaded instance which is the opposite of the idiomatic use of `operator get`.

PR #1440, opened 2026-05-23
1432 extract feature reminders module

ThomasKiljanczykDev:
@Futsch1 penultimate PR for the modularization effort.

This one create the :feature:reminders module.
I had to create additional :core:ui module to achieve correct module dependency graph.

PR #1441, opened 2026-05-23
1433 extract feature UI module

ThomasKiljanczykDev:
@Futsch1 final PR, merge after the PR #1440


feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/model/OverviewState.kt
```diff
@@ -0,0 +1,32 @@
...
+    SKIPPED,
+    LOCATION
+}
+
+fun OverviewState.toString(context: Context): String {
+    return when (this) {
+        OverviewState.PENDING -> context.getString(com.futsch1.medtimer.core.ui.R.string.please_wait)
+        OverviewState.TAKEN -> context.getString(com.futsch1.medtimer.core.ui.R.string.taken)
+        OverviewState.SKIPPED -> context.getString(com.futsch1.medtimer.core.ui.R.string.skipped)
+        OverviewState.RAISED -> context.getString(R.string.reminded)
```
Futsch1, line 19: Shouldn't this string also be part of core.ui? Same for the R.drawable.bell below?

app/src/main/java/com/futsch1/medtimer/OptionsMenu.kt
```diff
@@ -144,7 +146,7 @@ class OptionsMenu @AssistedInject constructor(
         val item = menu.findItem(R.id.settings)
         item.setOnMenuItemClickListener { _: MenuItem? ->
             try {
-                navController.navigate(R.id.action_global_preferencesFragment)
+                navController.navigate(com.futsch1.medtimer.feature.ui.R.id.action_global_preferencesFragment)
```
Futsch1, line 149: Here (and in maaaany other places): Why do you fully qualify the resources? Shouldn't it be just imported?

feature/reminders/src/main/res/values/strings.xml
Futsch1, line 1: Does it really make sense to separate the strings into feature.reminders and core.ui? Or is this required from the module logic/potential warnings?

Futsch1:
@ThomasKiljanczykDev I have some comments/questions to be resolved before merging

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/model/OverviewState.kt
ThomasKiljanczykDev: Both are only used in the `:feature:ui`.
The current strategy is to co-locate resources in a feature module.
If the resource is shared, it's promoted to the `:core:ui` module.

This reduces the numbers of rebuilds of the `:core:ui` module and other feature modules due to less frequent changes.

The rebuild impact should be moderate if we centralized the resources in the `:core:ui`.
I'll centralize the resources in the `:core:ui` to remove the need of having to decide where to put resources.

app/src/main/java/com/futsch1/medtimer/OptionsMenu.kt
ThomasKiljanczykDev: The idea for full-qualified path for `R` is to avoid confusion as to from which module the string comes from.
If we centralize the strings in the `:core:ui` module, there reason for the fully qualified path will be gone.

ThomasKiljanczykDev:
@Futsch1 I've addressed your comments

app/src/main/java/com/futsch1/medtimer/OptionsMenu.kt
ThomasKiljanczykDev: The app uses `nonTransitiveRClass`, so the FQN can be skipped only in modules that contain the string resources in them

ThomasKiljanczykDev:
@Futsch1 all green.

Do you have any more comments I need to address?

PR #1446, opened 2026-05-26
Introduce coding guidelines

ThomasKiljanczykDev:
@Futsch1 I've created coding guidelines to improve output quality for AI generated code.
The guidelines include Jetpack Compose to speed up the work on it.

What's in the PR:
- introduce coding guidelines
- rename `doc` directory to `docs` to align with standard naming in the industry


.github/copilot-instructions.md
Futsch1, line 1: From my point of view, you wouldn't need the instructions for Copilot specifically. GitHub Copilot will read AGENTS.md in every session as one of its first steps (as well as most other coding agents, and Claude Code would re-route the CLAUDE.md to the AGENTS.md), so this file is redundant from my point of view. As it should contain the same information, it would only put the same stuff into the content twice.

docs/guidelines/ai-assisted-development.md
```diff
@@ -0,0 +1,125 @@
...
+- **Dependency, AGP, Kotlin, KSP, or Hilt upgrades** — especially across majors.
+  These cascade into build issues, Kotlin metadata mismatches, and lint surprises.
+
+### Always
+
+- Run the gates above before pushing.
+- Use the project's commit format (see [Attribution and commits](#attribution-and-commits)).
+- Keep changes scoped to the task; surface unrelated problems separately rather than "fixing" them silently.
+
+## Privacy and data
```
Futsch1: This is partly redundant with the Never chapter.

docs/guidelines/ai-assisted-development.md
```diff
@@ -0,0 +1,125 @@
...
+- **Insecure output / hallucinations** — validate logic against the system, not just the type checker.
+  Lint and SonarQube are necessary but not sufficient.
+- **Sensitive-information disclosure** — watch for the model echoing context (file paths from your machine, environment values, tokens) into code or comments.
+- **Prompt injection** — be cautious when an agent processes untrusted content (issue text, fetched web pages); don't let it act on instructions embedded in
+  data.
+- **Supply-chain hygiene** — verify every suggested Gradle dependency: correct group/artifact (guard against typo-squats and hallucinated packages), maintained,
+  and pinned via `gradle/libs.versions.toml`.
+  New dependencies are an "Ask first" change (see above).
+
+## Attribution and commits
```
Futsch1: I don't think I want to be constrained by a commit message format. Same goes for branch names.

Reason behind is that I want to be able to do quick iterations sometimes on bugs I discover myself or refactorings I do. This might not align with community best practices, I know. Let's discuss it though.

docs/guidelines/ai-assisted-development.md
Futsch1, line 1: I do not quite understand the reason on why we need this file. It says it is directed at a human reader (I suppose?), but contains mostly the guidelines I would expect in the AGENTS.md - wouldn't this be enough?

docs/guidelines/documentation-guidelines.md
```diff
@@ -0,0 +1,94 @@
...
+- Keep relative links valid when reorganizing: after moving files, re-resolve every inbound link.
+
+## Maintenance and archiving
+
+- When a doc is superseded, **move it to `docs/archive/`** with a one-line pointer to its replacement — don't delete history, and don't leave stale guidance
+  live.
+- Date-stamp evidence docs and don't edit them after the fact; supersede instead.
+- **No secrets or real user data in docs** — same rule as code (see [ai-assisted-development.md](ai-assisted-development.md)).
+  Reference config keys by name, never paste real values.
+- Each guide carries a `_Last reviewed: YYYY-MM-DD._` footer; bump it when you revisit the content and re-confirm the linked sources.
```
Futsch1, line 86: I do not see why this would be helpful in a version controlled, PR reviewed document.

docs/guidelines/jetpack-compose.md
Futsch1, line 1: No opinions on this from my side - since I am not experienced with Compose. Trusting your judgement here.

docs/guidelines/testing.md
```diff
@@ -0,0 +1,158 @@
...
+Order of preference:
+
+1. **JVM unit tests** (`app/src/test/`, `*/src/test/`).
+   Fast, deterministic, no emulator, run on every push.
+   This is where most new tests should land.
+2. **JVM unit tests with Robolectric** when an Android framework class is genuinely needed (Context, Resources, Looper).
+   Still fast, still no emulator.
+3. **Instrumented tests** (`app/src/androidTest/`) as a last resort.
+   They're slow, flaky, and only run in CI (`compatibilityTest.yml` on API 28 and 36, plus `JacocoDebugCodeCoverage` in `test.yml`).
+   **Do not run instrumented tests locally** unless explicitly investigating a CI failure — they require an emulator and take significantly longer than JVM
```
Futsch1: I disagree. If you provide a new feature and create an Android instrumented test, both the AI and a human should absolutely run single tests. What should not be done is running the full test suite per default.

docs/guidelines/testing.md
```diff
@@ -0,0 +1,158 @@
...
+- Generated code.
+
+## Structure and naming
+
+- **Arrange / Act / Assert**, with the three sections visually separated by blank lines.
+- Test functions use **backtick names** describing the behavior:
+
+  ```kotlin
+  @Test
+  fun `should mark reminder as taken when user taps action`() { /* … */ }
```
Futsch1, line 73: I had tried this some time back, but it caused issues. I can no longer remember what it was, but are you sure this works?

Futsch1 (changes requested):

.github/copilot-instructions.md
ThomasKiljanczykDev: Github copilot instructions is needed when using Copilot from the Github page. This copilot is not reading AGENTS.md

docs/guidelines/ai-assisted-development.md
ThomasKiljanczykDev: This document is for AI assisted development, so my assumption is that the AI agent should adhere to this convention.

These are guidelines, not restrictions.
There's no strict enforcement and you can ignore them if needed

docs/guidelines/testing.md
ThomasKiljanczykDev: I can modify it so it recommends manually running the instrumented tests.
If an AI agent would run the tests, you would get a pretty long session and probably consume quite a bit of tokens.

EDIT: ok, narrowed to just the new tests, both human an AI should run the new tests

docs/guidelines/jetpack-compose.md
ThomasKiljanczykDev: This document will probably be modified in the first compose migration branch to adapt to the realities of this project.

docs/guidelines/documentation-guidelines.md
ThomasKiljanczykDev: The only thing this is useful is to give this information to the AI agent.
It's faster to read this information from the file, than call git

.github/copilot-instructions.md
ThomasKiljanczykDev: https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review#customizing-copilots-reviews-with-custom-instructions

docs/guidelines/testing.md
ThomasKiljanczykDev: Android tests do not support it. You'd need minSdk 30 for that.

We can either have a split convention for this so that unit tests use the naming with spaces and Android tests use camelCase naming.

Or we could use camelCase for all tests.

I'd be for the split convention, because there should be quite a bit more unit tests and the naming with spaces is more readable.

ThomasKiljanczykDev:
@Futsch1 I have addressed your comments and introduced changes.

The ai-assisted-guidelines are meant more for a larger organization.
Here you're just one guy with a couple of occasional contributors, so I folded what's valuable into the AGENTS.md and removed the file.

.github/copilot-instructions.md
Futsch1: A bit inconclusive I would say, as at another place (https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions) it says it does respect AGENTS.md. And both the CLI and the VSCode plugin do so too. So I suggest we try it without (to not maintain instructions twice) and if the feedback obviously doesn't follow the rules, we add it later.

docs/guidelines/testing.md
Futsch1: As I see it, tokens will not be consumed while waiting, only when investigating the test logs. This would be the same effort as when fixing compile issues in terms of tokens. So please change as you proposed.

docs/guidelines/testing.md
Futsch1: Ok, let's do it like that.

docs/guidelines/ai-assisted-development.md
ThomasKiljanczykDev: The file has been removed

docs/guidelines/testing.md
ThomasKiljanczykDev: Already did

docs/guidelines/ai-assisted-development.md
ThomasKiljanczykDev: I've consolidated these sections

docs/guidelines/testing.md
ThomasKiljanczykDev: Done

.github/copilot-instructions.md
ThomasKiljanczykDev: Removed the file.

Futsch1 (approved):

PR #1481, opened 2026-06-10
Use dedicated notification channel for out-of-stock reminders

### LINKED ISSUE #1479: Use a different notification category for out of stock reminder
Currently the out of stock reminders and medicine reminders are both sent with "Default" category. This means that it's not possible to use different system level notification settings for them.
The out of stock reminders require restocking, which isn't as fast and easy as just taking the medicine.
I prefer to keep such notifications about what needs to be done later and has lower urgency as Silent notifications on Android.

Use different notifications categories for different types of notifications: at least for medicine reminders and for the out of stock reminders.

stek29:
Fixes #1479

Create a ReminderChannel enum to separate notification channel identity from importance level. This allows out-of-stock reminders to use their own Android notification channel ("out_of_stock") instead of sharing the default medicine reminder channel, enabling independent system-level settings (sound, vibration, visibility).

- Add ReminderChannel enum with DEFAULT, HIGH, and OUT_OF_STOCK
- Refactor ReminderNotificationChannel to use ReminderChannel
- Update OutOfStockNotificationFactory to use OUT_OF_STOCK channel
- Add notification settings preference for out-of-stock channel

Note: Translations for notification_settings_out_of_stock are machine- generated and may require native speaker review.

### DIFF
--- a/core/domain/src/main/java/com/futsch1/medtimer/core/domain/model/Medicine.kt
+++ b/core/domain/src/main/java/com/futsch1/medtimer/core/domain/model/Medicine.kt
@@ -29,6 +29,8 @@ data class Medicine(
     }
 
     enum class ReminderChannel(val channelId: String, val importance: Int) {
+        // "3" and "4" are legacy channel IDs from when the app used importance.toString() directly.
+        // Keeping them for backward compatibility so users don't lose their notification settings.
         DEFAULT("3", NotificationManager.IMPORTANCE_DEFAULT),
         HIGH("4", NotificationManager.IMPORTANCE_HIGH),
         OUT_OF_STOCK("out_of_stock", NotificationManager.IMPORTANCE_DEFAULT)
--- a/core/domain/src/main/java/com/futsch1/medtimer/core/domain/model/Medicine.kt
+++ b/core/domain/src/main/java/com/futsch1/medtimer/core/domain/model/Medicine.kt
@@ -28,6 +28,12 @@ data class Medicine(
         HIGH(NotificationManager.IMPORTANCE_HIGH)
     }
 
+    enum class ReminderChannel(val channelId: String, val importance: Int) {
+        DEFAULT("3", NotificationManager.IMPORTANCE_DEFAULT),
+        HIGH("4", NotificationManager.IMPORTANCE_HIGH),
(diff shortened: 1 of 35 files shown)


stek29:
I have added the machine/llm translations because the lint was complaining about the untranslated strings.

Please tell me if I should remove them instead.

I also decided not to touch other kinds of notifications just yet (the expiration date reminders might need to be updated too, for example).

stek29:
I've rebased on top of main since the checks had a transient failure about emulator

Futsch1:
@stek29 
I had a quick look on your changes. While they look sound, I am wondering if it would not be better and more flexible to be able to override the notification channel per reminder. So we would still have default, high and high + alarm on medicine level, but we could override this per reminder. Would this still solve your use case?

If we decide against it, it would also be necessary to add a quick link to the Android notification settings for the new channel from the preferences menu.

Additionally, does the change keep the existing notification settings for default and high (are they using the same channelId)?

stek29:
@Futsch1 I'm not sure about overrides per reminder. In my use-case I just want to be able to separate actual medicine reminder notifications from low stock reminders, no matter which medicine they're about.

> If we decide against it, it would also be necessary to add a quick link to the Android notification settings for the new channel from the preferences menu.

That's already done, see https://github.com/Futsch1/medTimer/pull/1481/changes#diff-1fbf1a903b8a2d10f511340615886f06718a336c200c70059fb30b9adab956c4

> Additionally, does the change keep the existing notification settings for default and high (are they using the same channelId)?

Yes. Nothing changes for default and high in terms of behaviour and the actual notifications, only some refactoring to separate notification channel and "importance" more clearly. But it's not that clear from the current code.

Old code used `NotificationManager`'s `IMPORTANCE_DEFAULT` and `IMPORTANCE_HIGH` converted to strings, and new code just uses "3" and "4" as strings. I can change it to `IMPORTANCE_DEFAULT.toString()` and `IMPORTANCE_HIGH.toString()`, or just add comments clearly showing why those strings are used for channel IDs.

stek29:
@Futsch1 one of the checks has failed with an issue which seems unrelated to this PR. Maybe you can restart it?

Futsch1:
@stek29 Decided to take over your suggestion as is. Thank you for your contribution!

PR #1540, opened 2026-07-06
Fix simulating interval reminders that start from processed

Futsch1:


Futsch1:
Addresses #1543 partly

PR #1544, opened 2026-07-07
Claude/barcode entity data layer y9tt6d

stec314:


Futsch1:
Please describe the intention and scope of this PR: as it now also adds a WearOS companion app, I no longer see clearly what it wants to change. Additionally, it is getting bigger and bigger and thus very hard to review and test properly.

Futsch1:
You will especially have to separate the wear implementation from the barcode feature.

Also, do not increase the database scheme version on every change, just commit one version increment with all your changes.

PR #1590, opened 2026-07-20
test: add comprehensive unit tests for :core:datastore module

fpMiguel:
## Summary

Adds **92 JVM unit tests** to `:core:datastore` — the module previously had no direct
tests. Covers `PreferencesDataSource` (32), `PersistentDataDataSource` (48) and
`BackupMapper` (12) with edge-case coverage, and moves the existing datastore tests
from `:app` into the module where the code under test lives.

## Changes

| File | Change |
|------|--------|
| `core/datastore/build.gradle.kts` | Robolectric/JUnit test configuration + test dependencies |
| `gradle/libs.versions.toml` | Added `androidx.test:core` to the version catalog |
| `core/datastore/src/test/.../BackupMapperTest.kt` | 12 tests — `toSettingsBackup` mapping + `applyTo` integration |
| `core/datastore/src/test/.../PreferencesDataSourceTest.kt` | 32 tests — roundtrips, enum parsing, preference overrides, Flow emission |
| `core/datastore/src/test/.../PersistentDataDataSourceTest.kt` | 48 tests — setters, notification IDs, pending-snooze CRUD, data Flow |
| `app/src/test/.../location/` (deleted) | Removed — tests moved to `:core:datastore` where the code under test lives |

## Notes

- Test infrastructure: Robolectric + JUnit 4 + kotlinx-coroutines-test
- Per review feedback, default-value tests and tests that merely re-state Android
  `SharedPreferences` behavior were removed
- _Note: AI assistance was used in drafting this code and/or tests._

PR #1591, opened 2026-07-20
Fix number format exception crashes and backup directory handling in PreferencesDataSource

fpMiguel:
## Summary

Fixes **4 bugs** in `PreferencesDataSource.kt` that caused crashes or incorrect behaviour when stored preference values are invalid. These bugs were discovered and documented by the test suite in [PR #1590](https://github.com/Futsch1/medTimer/pull/1590).

### 1. NumberFormatException on invalid stored integers (3 locations)

`numberOfRepetitions`, `repeatDelay`, and `snoozeDuration` were parsed with `?.toInt()`, which throws `NumberFormatException` when the stored preference value is not a valid integer (e.g. corrupt data after a failed backup restore).

**Fix:** replaced with `?.toIntOrNull()` — returns `null` for invalid input, falling back to the default value.

### 2. Backup directory default produces wrong URI

`default.automaticBackupDirectory` is `null`, but the code called `.toString()` on it, producing the literal string `"null"`. This was then parsed by `.toUri()` → `Uri.parse("null")`, returning a non-null "null" URI instead of `null`.

**Fix:** changed the default to `null` directly.

### 3. Empty strings not handled before URI parsing (2 locations)

Both `alarmRingtone` and `automaticBackupDirectory` called `?.toUri()` directly on the stored string. An empty string `""` passes the null-safe call and produces a non-null empty URI, making it indistinguishable from "not set".

**Fix:** added `.takeIf { it.isNotEmpty() }` before `.toUri()`.

### 4. JsonSyntaxException crash on corrupt home location JSON

`getHomeLocation()` calls `gson.fromJson()` without any error handling. If the stored JSON is corrupt (e.g. partial write), the app crashes on every startup — this is a critical stability issue.

**Fix:** wrapped deserialization in try-catch, returning `null` on failure.

## Verification

- Compiles cleanly: ✅
- App builds (full + foss): ✅
- Existing app unit tests (257): ✅ all pass
- When merged with the test PR branch: all 141 datastore tests pass ✅

_Note: AI assistance was used in drafting this code and/or tests._

### DIFF
--- a/core/datastore/src/main/java/com/futsch1/medtimer/core/datastore/PreferencesDataSource.kt
+++ b/core/datastore/src/main/java/com/futsch1/medtimer/core/datastore/PreferencesDataSource.kt
@@ -67,7 +67,11 @@ class PreferencesDataSource @Inject constructor(
 
     private fun getHomeLocation(): HomeLocation? {
         val json = sharedPreferences.getString(HOME_LOCATION, null) ?: return null
-        return gson.fromJson(json, HomeLocation::class.java)
+        return try {
+            gson.fromJson(json, HomeLocation::class.java)
+        } catch (_: Exception) {
+            null
+        }
     }
 
     fun clearHomeLocation() {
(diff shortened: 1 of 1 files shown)


Futsch1 (approved):

core/datastore/src/main/java/com/futsch1/medtimer/core/datastore/PreferencesDataSource.kt
```diff
@@ -67,7 +67,11 @@ class PreferencesDataSource @Inject constructor(
 
     private fun getHomeLocation(): HomeLocation? {
         val json = sharedPreferences.getString(HOME_LOCATION, null) ?: return null
-        return gson.fromJson(json, HomeLocation::class.java)
+        return try {
```
Futsch1: Can't you use kotlin.runCatching here?

PR #1605, opened 2026-07-28
Feature/compose overview

ThomasKiljanczykDev:
@Futsch1 overview Compose migration is done.

I had to rework the whole test framework because it was pain in the ass to work with it.
Tests passed locally.

You should test this manually to confirm nothing has been missed.
I've seen some weird behavior with stock.
Might be a bug to fix.

I also took some stylistic decisions to make the medicine color not look this ugly on the card.

image

image


Futsch1:
@ThomasKiljanczykDev 
Oh my... This is a huge one. First of all I want to express that I am really grateful for the effort you put into this! I will test and review this thoroughly, but since it does not touch core reminder functionality, I am confident. It might take me a few days however until I am ready to merge it, so please be patient.

Please also don't mind me asking a few questions here and there to make sure that I understand your intentions and decisions, since in the end I am responsible for maintaining it later and need to know what I am working with.

ThomasKiljanczykDev:
@Futsch1 there were some behavior changes to the re-raise event.
You should take a look.

I don't know if I messed up anything with it and I don't want to look at this thing for the next 3 days.

.github/workflows/instrumentedTests.yml
```diff
@@ -95,8 +95,6 @@ jobs:
             adb shell input keyevent 3
             adb shell settings put secure long_press_timeout 1500
             adb uninstall com.futsch1.medtimer || true
-            adb shell settings put global auto_time 0
-            adb shell su 0 toybox date 0801160025
```
Futsch1, line 99: Why did you disable setting the time? I understand that it is desirable to have the tests run independent of local time, but not setting the time by default creates a source of flakiness that can be avoided. Doesn't this just add complexity to the instrumented tests?

Asserting when the time is not correct is tedious for local runs, but increasing complexity on the other hand also has a cost.

app/src/debug/AndroidManifest.xml
```diff
@@ -11,4 +11,10 @@
         android:name="android.permission.DUMP"
         tools:ignore="ProtectedPermissions" />
 
+    <application tools:ignore="MissingApplicationIcon">
+        <receiver
```
Futsch1: Why did you add this?

Futsch1:
@ThomasKiljanczykDev 
Started reviewing and testing and I have some immediate remarks (apart from the test failures, which you are surely aware of). I started the app on a Medium Phone emulator with API 36.1 and generated the test data. 

- Some reminder cards are way too big, see here:

image

- Can you add some space on the bottom (as in the medicine view), so that you can always scroll clear of the floating button?
- When you scroll to the bottom of the list and switch days, a medicine card is visible briefly on top of the days selector (hard to screenshot since it disappears immediately)
- The separator between medicine icons extends left of the filter buttons:

image

ThomasKiljanczykDev:
@Futsch1 I've fixed the tests and addressed the issues you reported so far.

docs/adr/0002-overview-scrollable-week-calendar.md
```diff
@@ -14,10 +14,22 @@ Future reminders require simulation (see ADR-0001). We maintain a prefetch buffe
...
 `WeekCalendarView.weekScrollListener` fires when the scroll settles on a new week. The listener auto-selects: today if the week contains today, otherwise the first day of the week. This keeps the overview always showing real data when scrolling, without requiring an explicit tap.
 
+## Week start
+
+Each visible week starts on the locale's culturally-correct first day of week (`WeekFields.of(locale).firstDayOfWeek`, via `rememberFirstDayOfWeek()`), matching
+the Statistics month calendar. Today lands wherever it falls in that week — it is not visually centered.
+
+An earlier version pinned `firstDayOfWeek` to `today - 3 days` so today always rendered in column 4 of 7, independent of locale. That produced a week grid with
+no real-world meaning (its boundaries didn't correspond to a calendar week in any locale), and it drifted daily since it was recomputed from `LocalDate.now()`
+on every composition. Locale-based weeks are stable and match user expectation.
```
Futsch1: I disagree with this choice. I consider the use case of "glimpsing to tomorrow's or yesterday's reminders" more important than a stable week selector that always starts at the same day. Please restore the previous behavior.

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/medicine/advancedReminderPreferences/AdvancedReminderSettingsActions.kt
```diff
@@ -0,0 +1,75 @@
...
+        linkedReminderHandlingFactory.create(reminder, fragment.lifecycleScope).deleteReminder(
+            fragment.requireContext(),
+            { NavHostFragment.findNavController(fragment).navigateUp() },
+            { }
+        )
+    }
+}
+
+@Composable
+fun AdvancedReminderSettingsMenu(actions: AdvancedReminderSettingsActions) {
```
Futsch1, line 86: I feel menus are a bit more inconsistent now (they were inconsistent before, it might have gotten a little worse). Sometimes they are buttons, sometimes labels behind a kebab icon, sometimes with icons, sometimes without.

While I do not have the perfect solution and this might be a future refactoring, I would prefer specifically this menu to be behind a kebab icon with text. The duplicate icon is not intuitive and will lead to an action without confirmation.

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/actions/ActionsVisitor.kt
```diff
@@ -162,8 +162,14 @@ class ActionsVisitor @Inject constructor(
         DeleteHelper.deleteItem(fragmentActivity, R.string.delete_re_raise_event, {
             fragmentActivity.lifecycleScope.launch {
                 undoStock(reminderEvent)
-                reminderEventRepository.delete(reminderEvent)
+                val raisedReminderEvent = reminderEvent.copy(
```
Futsch1: This implementation is the better choice, since it actually fixes a bug where an already raised reminder could not be properly rescheduled. And from what I can see it does not introduce any new issues.

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/ArcActionMenu.kt
Futsch1, line 1: There is no ready-to-use library for Compose to do this UI effect?

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/VariableAmountHandler.kt
```diff
@@ -29,33 +31,37 @@ class VariableAmountHandler @Inject constructor(
         val reminderNotificationData = intent.extras!!.toReminderNotificationData()
         if (!reminderNotificationData.valid) return
 
-        for (eventId in reminderNotificationData.reminderEventIds.reversed()) {
+        for (eventId in reminderNotificationData.reminderEventIds) {
```
Futsch1, line 34: Using `reversed()` here had a reason, but I cannot remember it. Have you investigated before changing?

Futsch1 (changes requested):

Futsch1:
@ThomasKiljanczykDev 
I did the review. I was not able to look into every change in detail, but I think I got the big picture. I have also done some testing, but did not find any new issues. Still, more manual testing might be required before releasing, but after addressing review comments and fixing tests I would want to merge the PR.

ThomasKiljanczykDev:
@Futsch1 after many flakiness mitigation + reschedule logic change revert, the tests passed.
It seems the branch is ready to merge

Futsch1:
@ThomasKiljanczykDev 
Great work! Would you address the review comments? Then I will merge.

.github/workflows/instrumentedTests.yml
ThomasKiljanczykDev: Re-added both lines.

The reason they went away is that the tests are now time-agnostic, so they no longer need them — the goal being that nobody has to pin their emulator to an arbitrary date just to run the suite locally. But you're right that pinning costs nothing on CI and removes the midnight-rollover class of flakiness for free, so it's back.

The stance stays "tests must not depend on wall-clock time"; the pinning is belt-and-braces on top of that, not a prerequisite.

---
*Generated by Claude Code. Comment to be reviewed by the replying developer.*

app/src/debug/AndroidManifest.xml
ThomasKiljanczykDev: It was a workaround for a build failure on the tablet test variant that appeared intermittently while I was reworking the test setup — not a deliberate behavioral change.

I've re-checked it: `assembleFullDebugAndroidTest` and `assembleFossDebugAndroidTest` both build clean without it, so the block is removed. ~If the failure resurfaces on CI I'll come back with the actual error rather than re-adding a blind override.~

---
*Generated by Claude Code. ~Comment to be reviewed by the replying developer~.*

EDIT: reviewed, the comment is correct

docs/adr/0002-overview-scrollable-week-calendar.md
ThomasKiljanczykDev: Restored. `firstDayOfWeek` is pinned to `today - 3 days` again, so today sits in column 4 of 7 and yesterday/tomorrow are always adjacent without scrolling. ADR-0002's *Week start* section now records that as the decision, with the locale-based variant as the rejected alternative.

One change from the original: the value is `remember`ed instead of being recomputed from `LocalDate.now()` on every composition. The old code shifted the whole grid by a day if the date rolled over while the screen was open; anchoring it once per composition lifetime keeps the grid still and lets today advance through it.

The unit tests that asserted locale-based week boundaries were replaced with one asserting today lands in column 4 regardless of locale, and the arrow tests now derive their expectations from the anchor instead of hard-coded dates.

---
*Generated by Claude Code. ~Comment to be reviewed by the replying developer.~*

EDIT: the comment is correct

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/medicine/advancedReminderPreferences/AdvancedReminderSettingsActions.kt
ThomasKiljanczykDev: Agreed, changed.
Both actions are now behind a kebab with text labels and leading icons.

No confirmation dialog on duplicate: it's non-destructive and undone by deleting, and the text label already makes the intent explicit.
There's a toast on success now to give some feedback to the user.
(A snackbar would be better, but the app has none yet and the destination screen is still View-based; worth its own change.)

If you'd still prefer a confirmation dialog, I'll add one.

Two follow-ups I'd rather not do in this PR:

- These actions belong on the reminder card, not behind "advanced settings". Delete already is there, as a long-press context menu — about as discoverable as the bare icon was. Left alone since the medicine editor is still View-based.
- Menu inconsistency: agreed. Scoped this to the one menu; a sweep across the other top bars deserves its own pass.


---
*Generated by Claude Code. ~Comment to be reviewed by the replying developer.~*

EDIT: i've trimmed the comment. These clankers are ever so over-verbose.

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/actions/ActionsVisitor.kt
ThomasKiljanczykDev: Heads-up: this code is no longer in the branch. I reverted it in `2011c2d6` while chasing instrumented-test flakiness, so the current implementation is the original `reminderEventRepository.delete(reminderEvent)` again.

I agree with your assessment, so I'd rather bring it back as its own PR with a dedicated test than smuggle a behavior change in under a "flakiness mitigation" commit — that way the rescheduling fix is reviewable on its own terms.

---
*Generated by Claude Code. Comment to be reviewed by the replying developer.*

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/ArcActionMenu.kt
ThomasKiljanczykDev: Not one that fits, no. The awkward part isn't the arc — it's that this menu fans out around an **arbitrary anchor inside a scrolling list** (the card's own status button, positioned via `positionOnScreen`), with RTL mirroring and screen-edge collision handling. Everything available assumes a fixed corner FAB.

What I looked at:

| Candidate | Why not |
|---|---|
| [`FloatingActionButtonMenu`](https://developer.android.com/reference/kotlin/androidx/compose/material3/package-summary#FloatingActionButtonMenu) (Material3 1.4.0-alpha) | The official option, and the one I'd prefer if it fit. It's a **vertical list** anchored to a `ToggleFloatingActionButton`, not an arc, and it can't be anchored to an arbitrary composable. |
| [JellyFab](https://github.com/iprashantpanwar/JellyFab) | Closest match — Compose, arc fan-out, scrim, Apache-2.0, no dependencies. But it's a corner-FAB menu: no arbitrary-anchor positioning, no RTL mirroring, no edge collision handling. Adding those means forking it, i.e. the same code plus a JitPack dependency. Also 52 stars / 20 commits / single author, which I'd weigh carefully for a component on the app's main screen. |
| [skydoves/compose-animations](https://github.com/skydoves/compose-animations) | Has a radial FAB that fans satellites along an arc with staggered springs — but it's a sample collection, not a published artifact, so it'd be copy-paste rather than a dependency. |

I skipped the View-based ones (MaterialArcMenu and similar) since interop would cost more than the current implementation.

Happy to revisit if `FloatingActionButtonMenu` gains arbitrary anchoring once it stabilises.

---
*Generated by Claude Code. ~Comment to be reviewed by the replying developer.~*

EDIT: reviewed. It's correct. As a clarification, I did guide the research process and checked some of these libraries before letting this comment through. I'm checking [skydoves/compose-animations](https://github.com/skydoves/compose-animations) right now if it does the animation better than we do

EDIT 2: skydover animations have nothing worth transfering into the repo. I stand by the answer that we need the hand-rolled implementation. The upside is that it handles a lot of edge cases now and probably won't be changed that much.

feature/ui/src/main/java/com/futsch1/medtimer/feature/ui/overview/VariableAmountHandler.kt
ThomasKiljanczykDev: Yes — the reason is recoverable from history, and it no longer applies.

You added `reversed()` in `25aa817e` ("Finish implementing variable amount"), in the same hunk that changed the dialog call to `activity.runOnUiThread { dialogHelper.show() }`.
That loop showed **every** dialog in a single pass without waiting for any of them, so they stacked as overlapping windows — and the last `show()` lands on top.
`reversed()` was what made the *first* reminder's dialog end up on top, i.e. the one the user saw first.

The loop now suspends on each dialog (`suspendCancellableCoroutine`) and only shows the next once the previous is dismissed.
Dialogs are therefore displayed strictly in list order, so `reversed()` would now literally invert the order the user works through them. Keeping it would have been the regression.

---
*Generated by Claude Code. ~Comment to be reviewed by the replying developer.~*

EDIT: reviewed, it's correct.

docs/adr/0002-overview-scrollable-week-calendar.md
ThomasKiljanczykDev: Although I restored the previous behavior, I would disagree with your disagreement.

The 'glimpsing tomorrow/yesterday' logic only holds when on the current day 'week'.
If you scroll past to the previous/next 'week' then you're shown a barely sensical order of days instead of the week with days in order per your locale .

ThomasKiljanczykDev:
@Futsch1 I've addressed the review comments and responded using Claude Code.
I've just finished final review of these comments to make sure they are correct.

The way I addressed some of the things is debatable and might require further changes.

EDIT: unfortunately, there are new flaky failures

Futsch1:
@ThomasKiljanczykDev 
The tests ran successfully once, but now fail again in both tablet variants (which seems to be more susceptible to flakiness since it requires more CPU due to higher resolution). Can you try to make them more robust? (I know they have not been 100% reliable before, but the latest versions before this PR was fairly stable). And I know that this is a pain, but I still consider this a major safety net for not introducing bugs into the app...

Futsch1 (approved):

Futsch1:
@ThomasKiljanczykDev 
Merged... Thanks a lot again for your contribution. I will release during the upcoming weekend I think and then closely monitor the feedback channels.

PR #1627, opened 2026-08-08
Use content padding for the overview event list

ThomasKiljanczykDev:
A quick fix to use contentPadding instead of a 'spacer item' in the overview event list


Futsch1:
I'll investigate the failed test in a separate branch (need to get used to the new instrumented test way).

Issue #1418, opened 2026-05-17
Battery dead = missing medications


marsianer:
**Describe the bug**
If the smartphone battery is dead, missing medications that were marked as automatically taken will not be added.

**Expected behavior**
Automatic addition of missing medications

PR #1648, opened 2026-08-14
#1418 Add regression tests for missed automatically taken reminders

fpMiguel:
## Summary

Regression tests for #1418 ("Battery dead = missing medications"). When the phone is
off at the time an automatically taken dose is due, that dose should still be recorded
as taken. The behavior already exists on `main` (fix `e657feb2`, December 2025, which
predates the issue). This PR adds tests so it cannot break unnoticed.

No production code changes.

## Changes

- **`MissedAutomaticallyTakenReminderTest`** (JVM / Robolectric, new): three scenarios.
  A missed automatically taken reminder is caught up and recorded as `TAKEN`. A missed
  reminder without that flag is raised instead. A reminder that fires right on time
  behaves correctly. This is the first JVM test for the catch-up path; nothing on `main`
  covered it before.
- **`MissedAutomaticallyTakenInstrumentedTest`** (instrumented, new): the scenario from
  the issue, end to end on a device. An automatically taken reminder with a past time is
  caught up and shown as taken in the overview, and no "taken" action stays in the
  notification shade. This complements `NotificationTest.automaticallyTakenTest`, which
  fires at a future time.
- **`TestReminderContext` fake fix**: the `create()` stub returned a `Long` where the
  repository interface declares `ReminderEvent`. Any test using that path crashed. No
  existing test hit it; the new one did.

## Testing

- JVM: `MissedAutomaticallyTakenReminderTest`, 3/3 green. Also verified from a clean
  checkout of `main` with the fake fix applied.
- Instrumented (Android 16 emulator, API 36): the new test passes 1/1, and the existing
  `NotificationTest.automaticallyTakenTest` passes 1/1 on the same code.
- Test data is synthetic ("Test med").

*AI was used to create this pull request*

### DIFF
--- a/app/src/androidTest/java/com/futsch1/medtimer/MissedAutomaticallyTakenInstrumentedTest.kt
+++ b/app/src/androidTest/java/com/futsch1/medtimer/MissedAutomaticallyTakenInstrumentedTest.kt
@@ -1,0 +1,45 @@
+package com.futsch1.medtimer
+
+import com.adevinta.android.barista.rule.flaky.AllowFlaky
+import com.futsch1.medtimer.core.ui.R
+import com.futsch1.medtimer.utilities.scheduleRemindersNow
+import dagger.hilt.android.testing.HiltAndroidTest
+import org.junit.Test
+
+/**
+ * Higher-level verification for https://github.com/Futsch1/medTimer/issues/1418
+ * ("Battery dead = missing medications").
+ *
+ * The #1418 scenario: the device is off (battery dead) when an automatically-taken
+ * reminder falls due. On the next app reschedule/start, the missed dose must be
+ * caught up and recorded as TAKEN - not left pending and not raised for a manual action.
+ *
+ * End-to-end on the device (frozen-clock harness): an automatically-taken reminder whose
+ * time is ALREADY BEHIND now ([MedTimerTestBase.earlierToday]) is processed via
+ * [scheduleRemindersNow]; the overview must show it as taken and no "taken" action may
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/com/futsch1/medtimer/MissedAutomaticallyTakenInstrumentedTest.kt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Higher-level verification for https://github.com/Futsch1/medTimer/issues/1418
+ * ("Battery dead = missing medications").
+ *
+ * The #1418 scenario: the device is off (battery dead) when an automatically-taken
+ * reminder falls due. On the next app reschedule/start, the missed dose must be
+ * caught up and recorded as TAKEN - not left pending and not raised for a manual action.
+ *
+ * End-to-end on the device (frozen-clock harness): an automatically-taken reminder whose
```
Futsch1, line 17: This does not nail the actual issue and code path which is taken when the device is off. In this case, a special routine is triggered that is called when the Android device is restarted. This cannot be tested via AndroidTest, since it would have to reboot the device (or I am not aware about how to do it). The test you added is no regression test for the referenced issue.

app/src/test/java/com/futsch1/medtimer/processortests/MissedAutomaticallyTakenReminderTest.kt
Futsch1, line 1: I think most of the cases tested here are already covered by other tests. I am sorry, but I do not see additional value in this test...

Futsch1 (changes requested):

PR #1653, opened 2026-08-14
#1494 Fix alarm displaying previous events when a second alarm arrives

### LINKED ISSUE #1494: Alarm can display previous events instead of current ones
**Describe the bug**
When an alarm is firing, it can display an alarm for the previous set of events instead of the current set of events that fired the alarm.

**To Reproduce**
1. Enable "Exact reminders" (for easier debugging), and "Combine same time notifications"
2. Add a medicine titled "test main", set its priority to High + Alarm, and add two reminders (one for 8:00 AM, and the other for 8:01 AM)
3. Duplicate "test main", rename to "test1", and delete the 8:01 AM reminder
4. Duplicate "test1", rename to "test2", and set the sole reminder to 8:01 AM
5. Lock the device, and wait until 8:00 AM
6. Press the home button to hide the alarm, and mark the "test main" and "test1" reminders as taken via the notifications
7. Lock the device, and wait until 8:01 AM
8. Notice that the alarm displays reminders for "test main" and "test1", _not_ "test main" and "test2"

**Expected behavior**
The alarm to display for the current set of events, not for the previous ones.

**Screenshots**

https://github.com/user-attachments/assets/427f0958-7d49-4d71-ac9e-67a5b9e8f590

**Smartphone**

- Device: Google Pixel 8a, Android Emulator
- Android version: Android 16

**Additional context**
I sometimes "silence" the alarm by pressing on the home button, to avoid marking the alarm as taken while I'm walking to take my meds, but without the alarm blaring until I do so.

fpMiguel:
## Summary

Fixes #1494. When a second alarm arrives while the alarm screen is already showing,
the screen kept displaying the previous alarm's events. The alarm activity is launched
with `NEW_TASK | SINGLE_TOP` (and `singleInstance` in the manifest), so a new alarm
while the screen is up is delivered via `onNewIntent`, which was never handled. The new
intent was dropped and the screen stayed on the old events.

No other behavior changes; the fix is confined to the alarm screen and its test.

## Changes

- **ReminderAlarmActivity**: handle `onNewIntent` by storing the new intent and
  replacing the displayed fragment with the new alarm's data.
- **AlarmFragment**: replacing the fragment no longer destroys the activity. The
  finish-and-remove-task behavior (which kept the alarm out of recents) is preserved
  for actual closes, including the system back button, via an `isFinishing` guard.
- **AlarmIntentRedeliveryTest** (new, instrumented): seeds two alarms, delivers the
  second alarm's intent through the real framework entry point
  (`Instrumentation.callActivityOnNewIntent`), and asserts the screen switches to the
  new alarm's medicine. Without the fix this test fails with "Alarm screen should show
  the second alarm's medicine (Meds B), got: Meds A" (verified against the unfixed
  parent commit).
- **RepositoryEntryPoint** (new, test-only): Hilt entry point for seeding test data
  directly, matching the existing entry-point pattern.

Note: open PR #1504 proposed a similar fix without a regression test and without the
fragment-lifecycle handling; this PR covers both.

## Testing

- Regression test: fails on the unfixed parent commit (RED, bug reproduced), passes on
  this branch (GREEN).
- Instrumented regressions: `NotificationTest` notificationTest, alarmTest,
  automaticallyTakenTest: 3/3 pass.
- JVM unit suite: 132 tests, 0 failures.
- Lint: clean.

*AI was used to create this pull request*

### DIFF
--- a/app/src/androidTest/java/com/futsch1/medtimer/AlarmIntentRedeliveryTest.kt
+++ b/app/src/androidTest/java/com/futsch1/medtimer/AlarmIntentRedeliveryTest.kt
@@ -1,0 +1,122 @@
+package com.futsch1.medtimer
+
+import android.app.Activity
+import android.widget.TextView
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.runner.lifecycle.ActivityLifecycleMonitorRegistry
+import androidx.test.runner.lifecycle.Stage
+import com.futsch1.medtimer.core.domain.model.Medicine
+import com.futsch1.medtimer.core.domain.model.Reminder
+import com.futsch1.medtimer.core.domain.model.ReminderEvent
+import com.futsch1.medtimer.feature.reminders.alarm.ReminderAlarmActivity
+import com.futsch1.medtimer.feature.reminders.api.notificationData.ReminderNotificationData
+import com.futsch1.medtimer.harness.RepositoryEntryPoint
+import com.futsch1.medtimer.utilities.pollUntil
+import dagger.hilt.android.EntryPointAccessors
+import dagger.hilt.android.testing.HiltAndroidRule
+import dagger.hilt.android.testing.HiltAndroidTest
+import kotlinx.coroutines.runBlocking
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.time.Instant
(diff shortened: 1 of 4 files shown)


Futsch1:
This is a tricky one. I also tried to fix it (see https://github.com/Futsch1/medTimer/pull/1504), but the behavior of Android differs between SDK 28 and SDK 36. So I guess the fix needs to check SDK versions and specifically find out which version changed the behavior.

